PR #2, opened 2019-10-17
Initial implementation of validation framework, with an example test for value transfer

anorth:
This is an initial sketch of the message validation framework. It's likely that a few things will need to change as we integrate into go-filecoin and lotus, but this proves out the basic idea.

At present, this implements the execution of a sequence of messages, but does not yet pay attention to those messages bundles into blocks, chain selection etc. Gas costs are still un-exercised.


anorth:
@frrist if this looks good to you, please squash and land it.

PR #3, opened 2019-10-17
message builder accepts gas params

frrist:
### Motivation
This PR plumb gas though the message factory. The benefit of this has yet to be fully realized since tests consuming this package do not produce messages that consume gas.

### DIFF
--- a/pkg/chain/message.go
+++ b/pkg/chain/message.go
@@ -44,7 +44,8 @@ const (
 // Integrations should implement this to provide a message value that will be accepted by the
 // validation engine.
 type MessageFactory interface {
-	MakeMessage(from, to state.Address, method MethodID, nonce uint64, value state.AttoFIL,
+	MakeMessage(from, to state.Address, method MethodID, nonce uint64,
+		value state.AttoFIL, gasPrice state.AttoFIL, gasLimit state.AttoFIL,
 		params ...interface{}) (interface{}, error)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/chain/message.go
```diff
@@ -44,7 +44,8 @@ const (
 // Integrations should implement this to provide a message value that will be accepted by the
 // validation engine.
 type MessageFactory interface {
-	MakeMessage(from, to state.Address, method MethodID, nonce uint64, value state.AttoFIL,
+	MakeMessage(from, to state.Address, method MethodID, nonce uint64,
+		value state.AttoFIL, gasPrice state.AttoFIL, gasLimit state.AttoFIL,
```
anorth: Gas limit should be in GasUnits

pkg/suites/example.go
```diff
@@ -16,6 +16,10 @@ import (
 func Example(t *testing.T, driver Driver) {
 	actors := make(map[state.Address]state.Actor)
 
+	// TODO this is poor UX, these will maybe need to go somehwere else?
+	gasPrice := big.NewInt(1)
+	gasLimit := big.NewInt(1000)
```
anorth: Two options (for a follow-up):
- Add state to the producer setting a gas price, and default gas limits in each of the sugar methods
- A functional options pattern to allow optional customization of the sugar methods behaviour

anorth (approved):
LGTM after GasUnits

PR #4, opened 2019-10-18
Single Message Application

frrist:
As discussed in person, we are considering single message application. The motivation here is to make assertions on the resulting state simpler.

To run this with go-filecoin you will need to checkout `chainval-single-msg`

### DIFF
--- a/pkg/chain/message.go
+++ b/pkg/chain/message.go
@@ -65,44 +65,44 @@ func NewMessageProducer(factory MessageFactory) *MessageProducer {
 	}
 }
 
-// Messages returns a slice containing all messages created by the producer.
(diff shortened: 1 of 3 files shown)


pkg/chain/message.go
```diff
@@ -53,7 +53,6 @@ type MessageFactory interface {
 // Actual message construction is delegated to a `MessageFactory`, and the message are opaque to the producer.
 type MessageProducer struct {
 	factory       MessageFactory
-	messages      []interface{}
```
anorth: We should keep this so the producer is still capable of exporting all the messages at the end (for comparison with other impls, replay, ...)

anorth (approved):
I'll restore the messages list and land this.

PR #5, opened 2019-10-18
add sugardrive and helper methods

frrist:
A `SugarDrive` is a wrapper around the `Driver` interface. The motivation for its creation was to make test code less verbose.

I have also modified the Factory NewState interface method to accept a slice to that its construction is deterministic.

### DIFF
--- a/pkg/suites/driver.go
+++ b/pkg/suites/driver.go
@@ -104,7 +104,7 @@ type SugarMessageProducer struct {
 	gasLimit state.GasUnit
 }
 
-func (s *SugarMessageProducer) Transfer(from, to *SugarActor, amount int64) state.Tree {
+func (s *SugarMessageProducer) MustTransfer(from, to *SugarActor, amount int64) state.Tree {
 	msg, err := s.producer.Transfer(from.address, to.address, big.NewInt(amount), s.gasPrice, s.gasLimit)
 	require.NoError(s.t, err)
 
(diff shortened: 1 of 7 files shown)


pkg/suites/driver.go
```diff
@@ -11,3 +15,119 @@ type Driver interface {
...
+	endState, msgReceipt, err := s.validator.ApplyMessage(s.exeCtx, s.tree, s.storage, msg)
+	require.NoError(s.t, err)
+	require.NotNil(s.t, endState)
+	require.NotNil(s.t, msgReceipt)
+
+	require.Equal(s.t, uint8(0), msgReceipt.ExitCode)
+	require.Equal(s.t, []byte{}, msgReceipt.ReturnValue)
+	require.Equal(s.t, state.GasUnit(0), msgReceipt.GasUsed)
+
+	newFromState, err := endState.Actor(from.address)
```
frrist: flagging that this method modifies the actor reference of the miner, from, and to actors- potentially an unpopular decision. From my perspective it makes writing code in Example less verbose so trying it out.

pkg/suites/driver.go
```diff
@@ -11,3 +15,132 @@ type Driver interface {
...
+	expectedFromBalance := prevFromBalance.Sub(&prevFromBalance, big.NewInt(amount))
+	actualFromBalance := big.Int(*newFromState.Balance())
+	require.Equal(s.t, expectedFromBalance.String(), actualFromBalance.String())
+
+	prevToBalance := big.Int(*to.actor.Balance())
+	expectedToBalance := prevToBalance.Add(&prevToBalance, big.NewInt(amount))
+	actualToBalance := big.Int(*newToState.Balance())
+	require.Equal(s.t, expectedToBalance.String(), actualToBalance.String())
+
+	// update actor state
```
frrist, line 140: flagging that this method modifies the actor reference of the miner, from, and to actors- potentially an unpopular decision. From my perspective it makes writing code in Example less verbose so trying it out.

frrist:
Closing in favor of #6

PR #6, opened 2019-10-21
Rework interfaces to provide a unified state wrapper, and state driver for manipulation/inspection

anorth:
This is inspired by the explorations in #5, but smaller scope and different enough that I did it separately.

Some follow-ons:
- have Validator hold onto the state (so it's not just an empty wrapper)
- functional options pattern for message producer gas price, defaults etc


frrist (approved):

pkg/suites/driver.go
```diff
@@ -1,13 +1,59 @@
...
+// AssertBalance checks an actor has an expected balance.
+func (d *StateDriver) AssertBalance(addr state.Address, expected uint64) {
+	actr, err := d.st.Actor(addr)
+	require.NoError(d.tb, err)
+	assert.Equal(d.tb, af(expected), actr.Balance())
+}
+
+// Helpers
+
+func af(v uint64) state.AttoFIL {
```
frrist, line 57: :pray:

PR #9, opened 2019-10-23
add actor code and singleton address enums

frrist:
### Motivation
Allow different implementations to use different actor codes and singleton actor addresses until spec converges on definitions

### Implementation.
- add SetSingletonActor method to Wrapper interface
- add FromSingletonAddress to message factory interface, required by the sugar methods.

### DIFF
--- a/pkg/state/actors.go
+++ b/pkg/state/actors.go
@@ -1,16 +1,16 @@
 package state
 
-type ActorCodeCid int
+type ActorCodeID int
 const (
-	AccountActorCodeCid  = ActorCodeCid(iota)
+	AccountActorCodeCid  = ActorCodeID(iota)
 	StorageMinerCodeCid
 	MultisigActorCodeCid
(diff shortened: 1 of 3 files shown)


pkg/state/actors.go
```diff
@@ -1,51 +1,17 @@
...
+	MultisigActorCodeCid
+	PaymentChannelActorCodeCid
 )
 
-// Builtin actor code CIDs.
-var (
-	InitActorCodeCid          cid.Cid
-	AccountActorCodeCid       cid.Cid
-	StorageMarketActorCodeCid cid.Cid
+type SingletonActorAddress int
```
anorth: Structurally fine, but it's a bit confusing to use "cid" and "address" for things that aren't. Rename these types to `ActorCodeId` and `SingletonActorId`.

anorth (approved):

PR #10, opened 2019-10-23
[wip] cbor: encode types to cbor array with example test

frrist:
This is a first pass at encoding types to CBOR. Packages that use this testing library should be able to encode messages parameters using `EncodeValues`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,9 +7,12 @@ require (
 	github.com/ipfs/go-bitswap v0.1.0 // indirect
 	github.com/ipfs/go-blockservice v0.0.2 // indirect
 	github.com/ipfs/go-cid v0.0.3
-	github.com/ipfs/go-ipld-cbor v0.0.2 // indirect
+	github.com/ipfs/go-ipld-cbor v0.0.2
 	github.com/ipfs/go-ipld-format v0.0.2 // indirect
 	github.com/ipfs/go-merkledag v0.0.2
+	github.com/libp2p/go-libp2p-core v0.0.2
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
+	github.com/multiformats/go-multihash v0.0.5
+	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.3.0
 )
--- a/go.mod
+++ b/go.mod
@@ -10,9 +10,7 @@ require (
 	github.com/ipfs/go-ipld-cbor v0.0.2
 	github.com/ipfs/go-ipld-format v0.0.2 // indirect
 	github.com/ipfs/go-merkledag v0.0.2
-	github.com/libp2p/go-libp2p-core v0.0.2
 	github.com/minio/blake2b-simd v0.0.0-20160723061019-3f5f724cb5b1
-	github.com/multiformats/go-multihash v0.0.5
 	github.com/pkg/errors v0.8.1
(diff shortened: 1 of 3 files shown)


pkg/state/types_test.go
```diff
@@ -0,0 +1,45 @@
+package state_test
+
+import (
+	"encoding/binary"
+	"math/big"
+	"testing"
+
+	"github.com/libp2p/go-libp2p-core/peer"
```
anorth: I'm striving to avoid depending on libp2p for this entire package. Since the peerid encoding is literally a no-op, depending on this is really unfortunate. Can we just use a manually-constructed byte array?

pkg/state/types.go
```diff
@@ -1,15 +1,47 @@
...
+		case PubKey:
+			arr = append(arr, v)
+		case PeerID:
+			arr = append(arr, v)
+		default:
+			return []byte{}, errors.Errorf("invalid type: %T at index: %d", p, i)
+		}
+	}
+	return cbor.DumpObject(arr)
+}
```
anorth, line 36: We need to separate the encoding of individual values from the encoding of the tuple/array, so that go-filecoin can access the former without the latter. Extract the switch statement to a function of its own.

pkg/state/types_test.go
```diff
@@ -0,0 +1,45 @@
...
+	bpid, err := RequireIntPeerID(t, 0).MarshalBinary()
+	require.NoError(t, err)
+
+	peerID := state.PeerID(bpid)
+
+	params := []interface{}{owner, sectorSize, peerID}
+
+	stuff, err := state.EncodeValues(params...)
+	assert.NoError(t, err)
+	assert.NotEmpty(t, stuff)
```
anorth: Without decoding, I have no idea if this does the right thing.

pkg/state/types_test.go
frrist: Yeah I can give that a shot

frrist: Errhh on second thought, this is just for convenience while testing- libp2p-peer isn't used outside of the test file, is that alright?

pkg/state/types.go
frrist: I have addressed this below with `EncodeValue`

pkg/state/types_test.go
anorth: No, that's the whole point. "just for convenience" isn't a good reason to pull in thousands of lines of code. I don't think anything here needs to assert that the byte array forms a valid peer id. it just needs to check that the same bytes come out as went in.

frrist: alright, addressed it in latest changes

go.mod
```diff
@@ -7,9 +7,12 @@ require (
 	github.com/ipfs/go-bitswap v0.1.0 // indirect
 	github.com/ipfs/go-blockservice v0.0.2 // indirect
 	github.com/ipfs/go-cid v0.0.3
-	github.com/ipfs/go-ipld-cbor v0.0.2 // indirect
+	github.com/ipfs/go-ipld-cbor v0.0.2
 	github.com/ipfs/go-ipld-format v0.0.2 // indirect
 	github.com/ipfs/go-merkledag v0.0.2
+	github.com/libp2p/go-libp2p-core v0.0.2
```
anorth: Please remove this line too

anorth (approved):

PR #12, opened 2019-10-31
Frrist/test with encoding

frrist:
### Motivation
This in conjunction with the changes in https://github.com/filecoin-project/lotus/pull/497 allow lotus to run the `AccountValueTransferTest` and `CreateStorageMinerAndUpdatePeerIDTest` by importing this package.

### Implementation
The key bit here was getting [`EncodeValue`](https://github.com/filecoin-project/chain-validation/compare/frrist/add-storage-power-address...frrist/test-with-encoding?expand=1#diff-114928b5b56bd96452914832f2804038R39) to produce a CBOR array for message method parameters that lotus could decode.


whyrusleeping (approved):
LMGTMTGM

PR #46, opened 2019-11-26
large refactor and new code gen

frrist:
### What
Implemented code gen similar to lotus and update most types to use its marshaling and unmarshaling methods, changes include:
- setup start of code gen. This will make interop with lotus easier and provides a path to filecoin as it is now using the same library.
- complete address implementation, I prefer it over casting everything as bytes.
- Replace AttoFIL and BytesAmount with BigInt type. Having less types feels simpler.
- Refactor directory structure


anorth:
This should be four separate PRs 😞. I can't provide feedback on this whole thing.

FYI Go-filecoin is not actually using cbor-gen (but probably will) so this won't be importable to go-filecoin at present.

frrist:
Closing in favor of #47, #48, #49, and #50

frrist:
The amount of overlap shared between go-filecoin, lotus, and the spec is fairly small, I think.  My thinking is that creating something that is bytes compatible with lotus will more useful right now than trying to fenagal something that works with both.

PR #72, opened 2020-02-12
gen: add code generation for message producer methods

frrist:


gen/gen.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	"github.com/filecoin-project/specs-actors/actors/builtin/multisig"
+	"github.com/filecoin-project/specs-actors/actors/builtin/paych"
+	"github.com/filecoin-project/specs-actors/actors/builtin/power"
+	"github.com/filecoin-project/specs-actors/actors/builtin/reward"
+)
+
+// This is a tool used to save some time writing all the message builder methods for actors.
+// It produces code that does NOT compile but saves a lot of typing.
+func main() {
```
frrist, line 23: a tool used to generate the methods in `chain/account.go`, `chain/cron.go`, `chain/multisig.go`, etc.

pkg/chain/message.go
```diff
@@ -1,326 +0,0 @@
-package chain
```
frrist, line 1: renamed to `pkg/chain/producer.go`

pkg/suites/driver.go
```diff
@@ -137,119 +137,3 @@ func (d *StateDriver) AssertMultisigState(multisigAddr address.Address, expected
 		assert.Contains(d.tb, multisig.Signers, e, fmt.Sprintf("expected Signer: %v, actual Signer: %v", e, multisig.Signers))
 	}
 }
-
-/*
```
frrist, line 141: removing these commented out methods as too much needs to change for them to be worth keeping around.

PR #99, opened 2020-03-04
fix typo in upstream method name

whyrusleeping:
depends on: https://github.com/filecoin-project/specs-actors/pull/229


frrist:
closed in https://github.com/filecoin-project/chain-validation/pull/101/

PR #104, opened 2020-03-07
Make tests spec-compliant and pass on Lotus 

arajasek:
**PR #113 should be merged first (or closed after this PR is merged)**

- provide internalCallSeqNum when calculating expected actor addresses
- use {idaddress, outer address} structs in expected message receipts
- use pubkey address, not ID address, to calculate expected message receipts
- expect txnId, not empty return, for multisig propose message

### DIFF
--- a/suites/message/create_actor.go
+++ b/suites/message/create_actor.go
@@ -109,15 +109,15 @@ func TestInitActorSequentialIDAddressCreate(t *testing.T, factory state.Factorie
 	var initialBal = abi_spec.NewTokenAmount(200_000_000_000)
 	var toSend = abi_spec.NewTokenAmount(10_000)
 
-	sender, senderID := td.NewAccountActor(drivers.SECP, initialBal)
+	sender, _ := td.NewAccountActor(drivers.SECP, initialBal)
 
 	receiver, receiverID := td.NewAccountActor(drivers.SECP, initialBal)
 
 	firstPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+1)
 	secondPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+2)
 
-	firstInitRet := td.ComputeInitActorExecReturn(senderID, 0, firstPaychAddr)
-	secondInitRet := td.ComputeInitActorExecReturn(senderID, 1, secondPaychAddr)
+	firstInitRet := td.ComputeInitActorExecReturn(sender, 0, 0, firstPaychAddr)
+	secondInitRet := td.ComputeInitActorExecReturn(sender, 1, 0, secondPaychAddr)
 
 	td.ApplyMessageExpectReceipt(
 		td.MessageProducer.CreatePaymentChannelActor(receiver, sender, chain.Value(toSend), chain.Nonce(0)),
--- a/suites/message/message_application.go
+++ b/suites/message/message_application.go
@@ -90,13 +90,13 @@ func TestMessageApplicationEdgecases(t *testing.T, factory state.Factories) {
 		}
 
 		// will create and send on payment channel
-		sender, senderID := td.NewAccountActor(drivers.SECP, initialBal)
+		sender, _ := td.NewAccountActor(drivers.SECP, initialBal)
 		// will be receiver on paych
 		receiver, receiverID := td.NewAccountActor(drivers.SECP, initialBal)
 
 		// the _expected_ address of the payment channel
 		paychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+1)
-		createRet := td.ComputeInitActorExecReturn(senderID, 0, paychAddr)
+		createRet := td.ComputeInitActorExecReturn(sender, 0, 0, paychAddr)
 		td.ApplyMessageExpectReceipt(
 			td.MessageProducer.CreatePaymentChannelActor(receiver, sender, chain.Value(toSend), chain.Nonce(0)),
 			types.MessageReceipt{ExitCode: exitcode.Ok, ReturnValue: chain.MustSerialize(&createRet), GasUsed: abi_spec.NewTokenAmount(1416)},
--- a/suites/message/multisig.go
+++ b/suites/message/multisig.go
@@ -38,7 +38,7 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
 		// expected address of the actor
 		multisigAddr := utils.NewIDAddr(t, 1+utils.IdFromAddress(aliceId))
 
-		createRet := td.ComputeInitActorExecReturn(aliceId, 0, multisigAddr)
+		createRet := td.ComputeInitActorExecReturn(alice, 0, 0, multisigAddr)
 		td.MustCreateAndVerifyMultisigActor(0, valueSend, multisigAddr, alice,
 			&multisig_spec.ConstructorParams{
 				Signers:               []address.Address{alice},
@@ -67,7 +67,7 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
 
 		multisigAddr := utils.NewIDAddr(t, 1+utils.IdFromAddress(outsiderId))
 
-		createRet := td.ComputeInitActorExecReturn(aliceId, 0, multisigAddr)
+		createRet := td.ComputeInitActorExecReturn(alice, 0, 0, multisigAddr)
 		// create the multisig actor
 		td.MustCreateAndVerifyMultisigActor(0, valueSend, multisigAddr, alice,
 			&multisig_spec.ConstructorParams{
@@ -144,6 +144,7 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
 
 		// Multisig actor address
 		multisigAddr := utils.NewIDAddr(t, 1+utils.IdFromAddress(outsiderId))
+		createRet := td.ComputeInitActorExecReturn(alice, 0, 0, multisigAddr)
 
 		// create the multisig actor
 		td.MustCreateAndVerifyMultisigActor(0, valueSend, multisigAddr, alice,
@@ -154,7 +155,7 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
 			},
 			types.MessageReceipt{
 				ExitCode:    exitcode_spec.Ok,
-				ReturnValue: multisigAddr.Bytes(),
+				ReturnValue: chain.MustSerialize(&createRet),
 				GasUsed:     big_spec.NewInt(1542),
 			})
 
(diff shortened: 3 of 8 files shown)


suites/message/create_actor.go
```diff
@@ -109,15 +109,15 @@ func TestInitActorSequentialIDAddressCreate(t *testing.T, factory state.Factorie
...
 
 	receiver, receiverID := td.NewAccountActor(drivers.SECP, initialBal)
 
 	firstPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+1)
 	secondPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+2)
 
-	firstInitRet := td.ComputeInitActorExecReturn(senderID, 0, firstPaychAddr)
-	secondInitRet := td.ComputeInitActorExecReturn(senderID, 1, secondPaychAddr)
+	firstInitRet := td.ComputeInitActorExecReturn(sender, 1, firstPaychAddr)
+	secondInitRet := td.ComputeInitActorExecReturn(sender, 2, secondPaychAddr)
```
frrist: I think this should continue to pass the ID based address here since it represents the runtimes immediate caller and is always expected to be an ID-address: https://filecoin-project.github.io/specs/#vmruntime-interface

arajasek: Thanks! I thought the spec was ambiguous here and picked the pubkey address (since that's what Lotus was doing), but it is very much not ambiguous. I've changed Lotus to do the correct thing, and undone my changes here :)

suites/message/create_actor.go
```diff
@@ -116,8 +116,8 @@ func TestInitActorSequentialIDAddressCreate(t *testing.T, factory state.Factorie
 	firstPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+1)
 	secondPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+2)
 
-	firstInitRet := td.ComputeInitActorExecReturn(senderID, 0, firstPaychAddr)
-	secondInitRet := td.ComputeInitActorExecReturn(senderID, 1, secondPaychAddr)
+	firstInitRet := td.ComputeInitActorExecReturn(senderID, 1, firstPaychAddr)
+	secondInitRet := td.ComputeInitActorExecReturn(senderID, 2, secondPaychAddr)
```
frrist: Sorry I didn't notice this change first review. I don't think this is right, `0` and `1` are the CallSeqNum of the from (`sender`) Actor. Actors are initialized with `CallSeqNum: 0` and `sender` shouldn't have had their `CallSeqNum` incremented at this point, unless I am missing something?

arajasek: If I'm understanding the spec correctly, they *should* have it incremented by this point. I'm basing that off the `ApplyMessage` function here: https://filecoin-project.github.io/specs/#vminterpreter-implementation

AFAIC tell, the rule is that you increment the nonce as soon as the basic validity checks have passed, and *before* any actual actor code is invoked.

Whether that *should* be the rule is ofc up for debate, but rn it seems that the spec says that the `CallSeqNum`s should already have been incremented.

anorth: That is what the spec code says, indeed. It wasn't consciously that way. [Discussion](https://filecoinproject.slack.com/archives/CHMNDCK9P/p1583875804246100)

suites/tipset/miner_flow.go
```diff
@@ -71,15 +71,15 @@ func TestMinerCreateProveCommitAndMissPoStChallengeWindow(t *testing.T, factory
 				minerActorID,
 				chain.Nonce(0), chain.Value(collateral),
 			),
-			types.MessageReceipt{ExitCode: exitcode_spec.Ok, ReturnValue: []byte{}, GasUsed: big_spec.Zero()},
+			types.MessageReceipt{ExitCode: exitcode_spec.Ok, ReturnValue: nil, GasUsed: big_spec.Zero()},
```
arajasek, line 74: Can someone confirm that this is correct? It looks like nil is what the specs actor returns, but there might be something I'm missing.

suites/tipset/miner_flow.go
```diff
@@ -42,7 +42,7 @@ func TestMinerCreateProveCommitAndMissPoStChallengeWindow(t *testing.T, factory
 	minerWorker, minerWorkerID := td.NewAccountActor(address.BLS, abi_spec.NewTokenAmount(1_000_000_000))
 	// The address of the miner actor
 	minerActorID := utils.NewIDAddr(t, utils.IdFromAddress(minerWorkerID)+1)
-	createMinerRet := td.ComputeInitActorExecReturn(builtin_spec.StoragePowerActorAddr, 0, minerActorID)
```
arajasek, line 45: pretty sure this was just wrong, but i might be missing something.

frrist: Since a storage miner actor is [created by](https://github.com/filecoin-project/specs-actors/blob/3f5510bf61303d1f629f8fce26ebe3194c911f93/actors/builtin/miner/miner_actor.go#L69-L70) the storage power actor I believe the StoragePowerActorAddr is the correct parameter here.

suites/tipset/miner_flow.go
frrist: Yes `nil` is the correct value. 

Note that lotus currently skips validation of the message receipt return field: https://github.com/filecoin-project/lotus/blob/testnet/3/chain/validation/factories.go#L44

suites/message/create_actor.go
```diff
@@ -109,15 +109,15 @@ func TestInitActorSequentialIDAddressCreate(t *testing.T, factory state.Factorie
...
 
 	receiver, receiverID := td.NewAccountActor(drivers.SECP, initialBal)
 
 	firstPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+1)
 	secondPaychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+2)
 
-	firstInitRet := td.ComputeInitActorExecReturn(senderID, 0, firstPaychAddr)
-	secondInitRet := td.ComputeInitActorExecReturn(senderID, 1, secondPaychAddr)
+	firstInitRet := td.ComputeInitActorExecReturn(sender, 0, 0, firstPaychAddr)
+	secondInitRet := td.ComputeInitActorExecReturn(sender, 1, 0, secondPaychAddr)
```
frrist, line 120: I believe the `senderID` address should be passed here since `ApplyMessage` will [resolve the sender](https://filecoin-project.github.io/specs/#vminterpreter-implementation) address through the InitActor's map before performing any operations.

Currently go-filecoin performs the operations as described in the spec here: https://github.com/filecoin-project/go-filecoin/blob/master/internal/pkg/vm/internal/vmcontext/vmcontext.go#L360

From a quick skim of the lotus code I don't believe this has been instrumented yet: https://github.com/filecoin-project/lotus/blob/testnet/3/chain/vm/vm.go#L252

suites/message/message_application.go
```diff
@@ -90,18 +91,21 @@ func TestMessageApplicationEdgecases(t *testing.T, factory state.Factories) {
...
 
 		// will create and send on payment channel
-		sender, senderID := td.NewAccountActor(drivers.SECP, initialBal)
+		sender, _ := td.NewAccountActor(drivers.SECP, initialBal)
 		// will be receiver on paych
 		receiver, receiverID := td.NewAccountActor(drivers.SECP, initialBal)
 
 		// the _expected_ address of the payment channel
 		paychAddr := utils.NewIDAddr(t, utils.IdFromAddress(receiverID)+1)
-		createRet := td.ComputeInitActorExecReturn(senderID, 0, paychAddr)
```
frrist, line 99: :arrow_up: ditto the comment about senderID address here

suites/message/message_application.go
```diff
@@ -116,5 +120,11 @@ func TestMessageApplicationEdgecases(t *testing.T, factory state.Factories) {
 			}, chain.Nonce(1), chain.Value(big_spec.Zero())),
 			types.MessageReceipt{ExitCode: exitcode.ErrIllegalArgument, ReturnValue: drivers.EmptyReturnValue, GasUsed: abi_spec.NewTokenAmount(1_000_000)},
 		)
+
+		// assert that the entire gas limit is deducted
+		gasReward := big.NewInt(td.MessageProducer.DefaultGasLimit())
+		gasReward = gasReward.Mul(gasReward, td.MessageProducer.DefaultGasPrice().Int)
+		senderBalance.Sub(senderBalance.Int, gasReward)
+		td.AssertBalance(sender, senderBalance)
```
frrist: Unsure this is the correct behaviour. The [ApplyMessage](https://filecoin-project.github.io/specs/#vminterpreter-implementation) section of the spec indicates that unused gas is to be refunded to the caller iff the sender address is valid (which it is here).

suites/message/multisig.go
```diff
@@ -38,7 +38,7 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
 		// expected address of the actor
 		multisigAddr := utils.NewIDAddr(t, 1+utils.IdFromAddress(aliceId))
 
-		createRet := td.ComputeInitActorExecReturn(aliceId, 0, multisigAddr)
```
frrist, line 41: :arrow_up: dittio comments about senderID address here.

suites/message/multisig.go
```diff
@@ -403,7 +406,7 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
 		var initialSigners = []address.Address{aliceId, bobId}
 
 		multisigAddr := utils.NewIDAddr(t, utils.IdFromAddress(chuckId)+1)
-
+		createRet := td.ComputeInitActorExecReturn(alice, 0, 0, multisigAddr)
```
frrist, line 409: senderID here as well.

frrist (changes requested):
Thanks for moving this forward, I think we need to have some more discussion here before this is ready to go in.

frrist (approved):
After reading https://filecoinproject.slack.com/archives/CHMNDCK9P/p1583985982257700 I understand that these changes are to be considered valid, the spec just doesn't reflect that yet. Let's hold on merging this for a little bit until some context landes in github.com/filecoin-project/specs. 

Thanks @arajasek for driving this forward.

suites/message/message_application.go
anorth: This change is invalid. Only the actually consumed gas should be spent.

If you believe there is disagreement about this, please remove this change so that this PR can otherwise land, and open an issue for discussion in specs.

anorth (changes requested):

arajasek:
@anorth I've rolled back the test on gas cost for reverts. I think Lotus might not be spec-compliant in its behaviour there, I'll look into it and add tests here later if necessary.

PR #105, opened 2020-03-09
implement gas constants for validation

frrist:
- Add code that applies a message and records the gas consumed by its application to a file. Gas values are keys on old state tree root, message cid, new state tree root. Allow this recording to be enabled/disabled with a boolean parameter.


frrist:
Closed in https://github.com/filecoin-project/chain-validation/pull/115

PR #106, opened 2020-03-09
add tests for invalid sender address:

frrist:
- send from address not in state tree
- send from address that is undef
- send from address with unknown protocol via staged message in multisig
actor


anorth:
We don't need to handle syntactically invalid messages (e.g. undef address) here.

PR #107, opened 2020-03-10
add test to check BLS -> SECP message ordering

frrist:


frrist:
This is a low priority test, will revisit after gas values settle.

PR #113, opened 2020-03-17
Add a new assertion to a test

arajasek:
The new assertion demonstrates a bug in Lotus behaviour during reverted messages.

### DIFF
--- a/suites/message/message_application.go
+++ b/suites/message/message_application.go
@@ -2,6 +2,7 @@ package message
 
 import (
 	"context"
+	"math/big"
 	"testing"
 
 	abi_spec "github.com/filecoin-project/specs-actors/actors/abi"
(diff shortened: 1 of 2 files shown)


suites/message/message_application.go
```diff
@@ -116,5 +120,11 @@ func TestMessageApplicationEdgecases(t *testing.T, factory state.Factories) {
 			}, chain.Nonce(1), chain.Value(big_spec.Zero())),
 			types.MessageReceipt{ExitCode: exitcode.ErrIllegalArgument, ReturnValue: drivers.EmptyReturnValue, GasUsed: abi_spec.NewTokenAmount(1_000_000)},
 		)
+
+		// assert that the entire gas limit is deducted
+		gasReward := big.NewInt(td.MessageProducer.DefaultGasLimit())
+		gasReward = gasReward.Mul(gasReward, td.MessageProducer.DefaultGasPrice().Int)
+		senderBalance.Sub(senderBalance.Int, gasReward)
+		td.AssertBalance(sender, senderBalance)
```
frrist, line 128: I left a comment about this here: https://github.com/filecoin-project/chain-validation/pull/104#discussion_r393815651. I would be in favor of closing this in favor of #104 since this appears to be duplicated from there.

arajasek: Yup, will do.

arajasek:
Will be merged as part of #104

anorth:
@frrist @arajasek for next time, separating changes into multiple PRs like this is the better approach. As it happens, this change is now the only thing blocking #104.

PR #115, opened 2020-03-18
generate gas validation constants from go-filecoin

frrist:
This PR does the following things:
- message and signed message serialization and CID methods
- GasMeter that tracks the amount of gas each test uses and writes it to a file
- GasConstantGenerate that generates a map keyed on state and message cid and value gas used for the application of the message to the state.
- ApplyMessageAndExpectReceipt returns the gas used for easier balance checking
- Exports the validator from TestDriver so all message application can have gas tracked
- a TestDriver `Complete` call that causes the GasMeter to write the values it tracked to a file.


gas_gen/gas_files/GASFILE_TestChainValidationMessageSuite_TestMultiSigActor_propose_and_approve
```diff
@@ -0,0 +1,5 @@
+bafy2bzaceczvqkzb6blvzohgecnw76bpyxmzbyc2o2u34veuouhip6sxwfsqm,bafy2bzaceb2bchdliyoznnpiboyjfqtj6hdyr245yutygfpjzxhj6ua42btve,bafy2bzaceczkiom6zwjsjpndjug76yfzk6o4aso5v4bcdmyubpee3szcel6lw,2037
```
frrist, line 1: one of the many files generated by the GasMeter

gas_gen/gas/constants.go
```diff
@@ -0,0 +1,50 @@
+package gas
+
+var GasConstants = map[string]int64{
```
frrist, line 3: This file is generated using all of the files in ./gas_files/

gas_gen/gen.go
```diff
@@ -0,0 +1,131 @@
...
+		defer codeFile.Close()
+	}
+
+	_, err = codeFile.WriteString(code)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func generateGoCodeFromGasFiles(gasFiles []string) (string, error) {
```
frrist, line 66: code that generates the map of state-message -> gas used

PR #121, opened 2020-03-20
refactor gas meter

frrist:
(taken from `Usage.md` in this PR)
# Usage

The gasmeter package is used to record and validate an implementations gas usage.

When recording, gasmeter will produce a file for each test that is ran. The file will be named after the test it is
generate for with all character but alpha numeric striped. The file will contain gas data in the form:
```text
GasUnitsUsed
GasUnitsUsed
GasUnitsUsed
...
...
```

When validating, gasmeter will search for the gas vale corresponding to the test it is running (ignoring non alpha numerica character)
in box/box.go. The order the messages are applied in correspond to the line number in the file. For example, the first
call to apply message in a test will validate against the first line in the file, second call to apply message to 
second line in file, and so on.
The same pattern is used for TipSet application.

## How To Record
1. Set the environment variable `CHAIN_VALIDATION_DATA` to the location of the chain-validation gas resources file. For 
most users this will be: `$GOPATH/chain-validation/box/resources`.
2. Set `RecordGas` in the implementations `ValidationConfig` to `true`.
3. Run tests you wish to record gas for, and verify files with names corresponding to the tests previously ran exist in
`$GOPATH/chain-validation/box/resources`
4. Run `make gen-gas` to generate `box/blob.go`.


drivers/gas_meter.go
```diff
@@ -1,100 +0,0 @@
-package drivers
```
frrist, line 1: file was moved.

drivers/test.go
```diff
@@ -242,16 +243,17 @@ func (b *TestDriverBuilder) Build(t testing.TB) *TestDriver {
...
 		StateDriver:     sd,
 		MessageProducer: producer,
 		validator:       validator,
 		ExeCtx:          exeCtx,
 
-		Config: b.factory.NewValidationConfig(),
+		Config: vconfig,
 
-		GasMeter: NewGasMeter(t, Validate),
+		GasMeter: gasmeter.NewGasMeter(t, vconfig.RecordGas()),
```
frrist: allow an implementation to configure if gas values are recorded.

gas_gen/gen.go
```diff
@@ -1,131 +0,0 @@
-package main
```
frrist, line 1: deleted code gen in favor of env var.

anorth:
> This change will require implementations using this package to make the gas files available during CI runs

I don't quite know how to interpret this. We need the tests to "just work" on CI, and cannot assume that the chain-validation package is checked out locally. Given the ~impossibility of loading resource files from a go module, this sounds like a blocker.

frrist:
> We need the tests to "just work" on CI, and cannot assume that the chain-validation package is checked out locally

Agreed. I think one way to address that is by making chain-validation a submodule in the projects using it, another way would be to go back to code generation.

anorth:
I could be convinced of the submodule approach. It may be worth a few minutes playing with https://godoc.org/golang.org/x/tools/go/packages#Package first to see if it provides a viable alternative.

box/resources/TestChainValidationMessageSuiteTestInitActorSequentialIDAddressCreate
```diff
@@ -1,2 +1,2 @@
-bafy2bzacecu7n7wbtogznrtuuvf73dsz7wasgyneqasksdblxupnyovmtwxxu,bafy2bzacebsjmkbe6lrpotwhk6zi7j7zllystyezo7fjerfqc5qorb5ziziv2,bafy2bzaceaxbu5r2sdn3tkaukzqya3m4y7bcrguohyzt7zqlxjtewlp4d5jo2,1918
```
frrist: regenerated these files and go-filecoin has made some fixes since the previous generation.

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Given a testing T, load the gas file associated with it and return a slice of the gas used by the test
+// an index in the slice represents the order of apply message calls.
+func LoadGasForTest(t testing.TB) []int64 {
+	f, found := box.Get(filenameFromTest(t))
+	if !found {
+		t.Fatal("can't find file")
+	}
+
```
frrist: Where we read the generated file

box/box.go
```diff
@@ -0,0 +1,52 @@
+//go:generate go run gen.go
+
+// Source: https://levelup.gitconnected.com/how-i-embedded-resources-in-go-514b72f6ef0a
```
anorth: Reference the repo at https://github.com/wso2/product-apim-tooling/tree/master/import-export-cli (pointed to by the blog post). This repo carries an Apache LICENSE file which thus permits this re-use.

```
Based on code from https://github.com/wso2/product-apim-tooling/tree/6292e88350868ddaa17ec58501672f008639dd5a/import-export-cli on 2020-03-21 under the Apache License 2.0.
```

drivers/test.go
anorth: Redundant after changes I landed recently.

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,160 @@
...
+		tracker:          list.New(),
+		T:                t,
+		record:           record,
+		gasIdx:           0,
+		expectedGasUnits: gasUnits,
+	}
+}
+
+func (gm *GasMeter) Track(oldState, newState cid.Cid, msg *types.Message, receipt types.MessageReceipt) {
+	if gm.record {
```
anorth: I removed this mode from the old tracker, it's not necessary. Please apply analogous changes here.

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,160 @@
...
+		gm.tracker.PushBack(&trackerElement{
+			oldState: oldState,
+			newState: newState,
+			msg:      msg,
+			receipt:  receipt,
+		})
+	}
+}
+
+func (gm *GasMeter) ExpectedGasUnit() int64 {
```
anorth: Return whether the unit was found so the caller can decide whether to check against it or not.

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+}
+
+// Given a testing T, load the gas file associated with it and return a slice of the gas used by the test
+// an index in the slice represents the order of apply message calls.
+func LoadGasForTest(t testing.TB) []int64 {
+	f, found := box.Get(filenameFromTest(t))
+	if !found {
+		t.Fatalf("can't find file: %s", filenameFromTest(t))
```
anorth: I haven't fully grokked the process, but this looks like it could make writing new tests painful.

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,160 @@
...
+			msg:      msg,
+			receipt:  receipt,
+		})
+	}
+}
+
+func (gm *GasMeter) ExpectedGasUnit() int64 {
+	if !gm.record {
+		defer func() { gm.gasIdx += 1 }()
+		return gm.expectedGasUnits[gm.gasIdx]
```
anorth: Check for index out of bounds.

anorth (commented):
Slightly hard to judge until the Usage.md describes the new full process.

I expect the gen_gas files to not be checked in since they are unused at runtime. They're only an intermediate artefact of generating new fixture data.

If as an impl I run with record=true can I expect the subsequent run with record=false to automatically pass because it's using new values now? What else needs to happen?

What do I need to do when writing a new test?

box/resources/TestChainValidationMessageSuiteTestInitActorSequentialIDAddressCreate
anorth: Please revert these or make a different PR that changes these values independent of big changes to the process that generates them.

Makefile
```diff
@@ -34,6 +34,9 @@ clean:
 gen:
 	$(GOBIN) run ./pkg/gen/main.go
 
+gen-gas:
+	$(GOBIN) generate ./box/...
+
```
frrist, line 39: after running an implementation in record mode this is the command a user runs to generate blog.go that adds gas values to box storage: https://github.com/filecoin-project/chain-validation/pull/121/files#diff-a9ff472173cbe0a48939f2d59c3412e4R8

box/resources/TestChainValidationMessageSuiteTestInitActorSequentialIDAddressCreate
frrist: the format of these files has changes such that they only contain gas values. I have removed them since they are not required. I would prefer to change the files and the process for generating them all at once.

gasmeter/gas_meter.go
frrist: I have removed the fatal call and now just log in the event they are not found. Will make new test writing easier.

box/resources/TestChainValidationMessageSuiteTestInitActorSequentialIDAddressCreate
anorth: If the format has changed, sure.

Makefile
anorth: Fold it into `gen`? Too many commands to remember otherwise.

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func (gm *GasMeter) Track(receipt types.MessageReceipt) {
+	gm.tracker.PushBack(&trackerElement{
+		receipt: receipt,
+	})
+}
+
+func (gm *GasMeter) ExpectedGasUnit() (int64, bool) {
```
anorth: Make the iterator-like functionality more clear with a name like `NextExpectedGas`

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,143 @@
...
+	defer func() { gm.gasIdx += 1 }()
+	if gm.gasIdx > len(gm.expectedGasUnits)-1 {
+		// didn't find any gas
+		return 0, false
+	}
+	return gm.expectedGasUnits[gm.gasIdx], true
+}
+
+// write the contents of gm.tracker to a file using the format:
+// oldStateCID,msgCID,newStateCID,GasUnits
```
anorth: Update comments

gasmeter/gas_meter.go
```diff
@@ -0,0 +1,143 @@
...
+func getTestDataFilePath(t testing.TB) string {
+	dataPath := os.Getenv(ValidationDataEnvVar)
+	if dataPath == "" {
+		t.Fatalf("failed to find validation data path, make sure %s is set", ValidationDataEnvVar)
+	}
+	return filepath.Join(dataPath, filenameFromTest(t))
+}
+
+// given a line of the form:
+// oldState,messageCid,newState,gasUnits
```
anorth: Update comments

anorth (approved):

Makefile
frrist: So `gen` doesn't actually generate anything, it prints the message producer boiler plate to the console, a developer needs to copy-paste that output into the correct files and make a few minor changes (method numbers). It was meant to serve as a tool to help me keep up with specs-actors changes when things were moving around a lot and I didn't sink a lot of time into making it ergonomic. I don't think remembering 2 different gen commands is unreasonable, if they are documented.

PR #122, opened 2020-03-21
Fix multisig to send from account actors only, exercise recursive send

anorth:
The prior multisig tests attempted to send top-level messages from the multisig actor itself (in order to satisfy caller validation), which is illegal. This changes to propose a protected transaction to the multisig via itself, which is the expected use case.

This exercises recursive message sends.

@frrist informs me that the hardcoded gas values in these tests are no longer used (even if when gas validation is turned on), because there's a new fixture-based gas validation mechanism. I zero'd them so as to avoid confusion.

FYI @frrist @whyrusleeping

### DIFF
--- a/chain/message_producer.go
+++ b/chain/message_producer.go
@@ -19,6 +19,7 @@ type MessageProducer struct {
 func NewMessageProducer(defaultGasLimit int64, defaultGasPrice big_spec.Int) *MessageProducer {
 	return &MessageProducer{
 		defaults: msgOpts{
+			value:    big_spec.Zero(),
 			gasLimit: defaultGasLimit,
 			gasPrice: defaultGasPrice,
 		},
--- a/drivers/test.go
+++ b/drivers/test.go
@@ -192,7 +192,7 @@ type TestDriverBuilder struct {
 
 	actorStates []ActorState
 
-	defaultGasPrice big_spec.Int
+	defaultGasPrice abi_spec.TokenAmount
 	defaultGasLimit int64
 }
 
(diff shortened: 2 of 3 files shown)


suites/message/multisig.go
```diff
@@ -220,258 +223,79 @@ func TestMultiSigActor(t *testing.T, factory state.Factories) {
...
-			NumApprovalsThreshold: initialNumApprovals + 1,
-			NextTxnID:             0,
-			InitialBalance:        valueSend,
-			StartEpoch:            td.ExeCtx.Epoch,
-			UnlockDuration:        unlockDuration,
-		})
-
-	})
-
-	t.Run("remove signer and decreases threshold", func(t *testing.T) {
```
anorth, line 311: The diff is weird, but I deleted tests below the "add signer" one. They are not exercising any novel VM or message execution behaviour, only the multisig actor itself. To the extent they did anything that the actor unit tests don't do, they should move there.

chain/message_producer.go
```diff
@@ -19,6 +19,7 @@ type MessageProducer struct {
 func NewMessageProducer(defaultGasLimit int64, defaultGasPrice big_spec.Int) *MessageProducer {
 	return &MessageProducer{
 		defaults: msgOpts{
+			value:    big_spec.Zero(),
 			gasLimit: defaultGasLimit,
```
frrist, line 23: Good idea! Wish I'd have done this a while ago :man_facepalming:

drivers/test.go
```diff
@@ -302,6 +302,12 @@ func (td *TestDriver) ApplyMessageExpectReceipt(msg *types.Message, expected typ
 	return expectedGasUsed
 }
 
+func (td *TestDriver) GetBalance(addr address.Address) big_spec.Int {
```
frrist: Why not return a token amount here?

frrist (approved):
Nice, lets get this in before: https://github.com/filecoin-project/chain-validation/pull/123

PR #126, opened 2020-03-22
Fix and clean up internal message send tipset test

anorth:
I was debugging the internal message send tipset tests. I fixed some things wrong with the test, but still fails on go-filecoin. 

There's nothing inherently "tipset" about that test so I've set up a new message test file intending to replace it with vanilla message tests. But this cleanup and partial fix can go in now.

### DIFF
--- a/drivers/test.go
+++ b/drivers/test.go
@@ -279,6 +279,13 @@ func (td *TestDriver) Complete() {
 	//td.GasMeter.Record()
 }
 
+func (td *TestDriver) ApplyMessage(msg *types.Message) types.MessageReceipt {
+	// TODO for failure cases we should consider catching panics here else they appear in the test output and obfuscate successful tests.
+	receipt, err := td.validator.ApplyMessage(td.ExeCtx, td.State(), msg)
+	require.NoError(td.T, err)
+	return receipt
+}
+
 func (td *TestDriver) ApplyOk(msg *types.Message) int64 {
 	return td.ApplyExpect(msg, nil)
 }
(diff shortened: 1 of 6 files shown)


drivers/test.go
```diff
@@ -328,15 +333,15 @@ func (td *TestDriver) AssertBalanceCallback(addr address.Address, thing func(act
...
-	}
 	if td.Config.ValidateExitCode() {
 		assert.Equal(td.T, expected.ExitCode, actual.ExitCode, "Expected ExitCode: %s Actual ExitCode: %s", expected.ExitCode.Error(), actual.ExitCode.Error())
 	}
 	if td.Config.ValidateReturnValue() {
 		assert.Equal(td.T, expected.ReturnValue, actual.ReturnValue, "Expected ReturnValue: %v Actual ReturnValue: %v", expected.ReturnValue, actual.ReturnValue)
 	}
+	if td.Config.ValidateGas() {
+		assert.Equal(td.T, expected.GasUsed, actual.GasUsed, "Expected GasUsed: %s Actual GasUsed: %s", expected.GasUsed.String(), actual.GasUsed.String())
+	}
```
anorth: Moved gas checking to last as "least important".

suites/tipset/internal_message_failure.go
```diff
@@ -25,88 +24,70 @@ import (
...
 		blkBuilder := drivers.NewTipSetMessageBuilder(td)
 
-		_, aliceId := td.NewAccountActor(drivers.SECP, abi_spec.NewTokenAmount(2_000_000_000_000))
-
-		// this address should not exist since the propose is sent to the wrong account actor method
-		nobody := utils.NewSECP256K1Addr(t, "pubbub")
+		alice, aliceId := td.NewAccountActor(drivers.SECP, abi.NewTokenAmount(2_000_000_000_000))
 
 		multisigAddr := utils.NewIDAddr(t, utils.IdFromAddress(aliceId)+1)
+		createRet := td.ComputeInitActorExecReturn(alice, 0, 0, multisigAddr)
```
anorth, line 37: Previously wrongly used `aliceId` and count of `1`.

chain/multisig.go
```diff
@@ -8,35 +8,35 @@ import (
 	"github.com/filecoin-project/chain-validation/chain/types"
 )
```
frrist: These files are generated, if this is a change we want then I propose we regenerate/change them all for consistency (I  volunteer to make that change if so). Is this change necessary? Seems like its more about style.

frrist (approved):
I don't see the need for the change in multisig message produce, otherwise LGTM.

chain/multisig.go
anorth: Oh, I didn't realise it was generated. Ok I'll revert, but yes could you please
- change the codegen so as not to copy params objects, but pass a pointer
- include a comment saying this is generated code and how to re-gen it

PR #129, opened 2020-03-22
Expand tests on internal message sends via multisig

anorth:
There are more to come, but go-filecoin already fails a bunch. These are all exercisable directly by use of the current multisig actor. I've left comments describing some more we could test with a custom actor (or through unknown vulns in our existing actors).

The tipset message failure tests are subsumed by these new ones.


suites/message/nested.go
```diff
@@ -33,61 +35,235 @@ func TestNestedSends(t *testing.T, factory state.Factories) {
...
+		balanceBefore := td.GetBalance(stage.creator)
+
+		// Multisig sends to new address and invokes pubkey method at the same time.
+		newAddr := td.Wallet().NewSECP256k1AccountAddress()
+		amtSent := abi.NewTokenAmount(1)
+		recpt := stage.send(newAddr, amtSent, builtin.MethodsAccount.PubkeyAddress, nil, 1)
+		assert.Equal(t, exitcode.Ok, recpt.ExitCode)
+		expected := bytes.Buffer{}
+		require.NoError(t, newAddr.MarshalCBOR(&expected))
+		assert.Equal(t, expected.Bytes(), recpt.ReturnValue)
```
arajasek: How does this work? My understanding is that the result of calling MethodsAccount.PubkeyAddress will be given to the multisig actor's Propose() method. However, Propose() will return nil, and so this address won't actually get put into the message receipt. Is that wrong?

suites/message/nested.go
```diff
@@ -33,61 +35,235 @@ func TestNestedSends(t *testing.T, factory state.Factories) {
...
+		td := builder.Build(t)
+		defer td.Complete()
+
+		stage := prepareStage(td, acctDefaultBalance, multisigBalance)
+
+		newAddr := td.Wallet().NewSECP256k1AccountAddress()
+		amtSent := abi.NewTokenAmount(1)
+		params := typegen.Deferred{Raw: []byte{1, 2, 3, 4}}
+		recpt := stage.send(newAddr, amtSent, builtin.MethodSend, &params, 1)
+		assert.Equal(t, exitcode.Ok, recpt.ExitCode)
```
whyrusleeping: Asserting the execution is fine here (Which i agree with) seems to be at odds with asserting the target actor doesnt get created down below on line 230

whyrusleeping: Ah, I think i get what youre asserting here. Invalid parameters should cause the internal send to fail.
I think that `MethodSend` should not care at all about the format of its parameters, the only reason we care that things are CBOR is because we happen to choose cbor as the encoding for our actors (an actor implementation should be able to choose whatever encoding they like, technically). And therefore, unless the receiver method tries to decode parameters, it shouldnt care whats in that field. (this can be used by payment processors to tag transactions, among other things)

suites/message/nested.go
frrist: @arajasek this reflects my understanding as well, the Multisig `Send` ignores any return value: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/multisig/multisig_actor.go#L311

anorth: You are quite right, thank you.

We should later change `Approve` to return the inner return value (but then to check that we'll have to use signers >1).

suites/message/nested.go
anorth: Ok. I'll change the test to assert successful transfer instead.

frrist (approved):

whyrusleeping (approved):

PR #130, opened 2020-03-22
ensure origin address is pubkey for actor creation

whyrusleeping:
This was causing some trouble for me

### DIFF
--- a/drivers/state.go
+++ b/drivers/state.go
@@ -93,9 +93,9 @@ func (d *StateDriver) newMinerAccountActor() address.Address {
 
 	// creat a miner, owner, and its worker
 	_, minerOwnerID := d.NewAccountActor(address.SECP256K1, big_spec.NewInt(1_000_000_000))
-	_, minerWorkerID := d.NewAccountActor(address.BLS, big_spec.Zero())
+	minerWorkerPk, minerWorkerID := d.NewAccountActor(address.BLS, big_spec.Zero())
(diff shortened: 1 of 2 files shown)


drivers/state.go
```diff
@@ -93,9 +93,9 @@ func (d *StateDriver) newMinerAccountActor() address.Address {
 
 	// creat a miner, owner, and its worker
 	_, minerOwnerID := d.NewAccountActor(address.SECP256K1, big_spec.NewInt(1_000_000_000))
-	_, minerWorkerID := d.NewAccountActor(address.BLS, big_spec.Zero())
+	minerWorkerPk, minerWorkerID := d.NewAccountActor(address.BLS, big_spec.Zero())
 	expectedMinerActorIDAddress := utils.NewIDAddr(d.tb, utils.IdFromAddress(minerWorkerID)+1)
-	minerActorAddrs := computeInitActorExecReturn(d.tb, builtin_spec.StoragePowerActorAddr, 0, 1, expectedMinerActorIDAddress)
+	minerActorAddrs := computeInitActorExecReturn(d.tb, minerWorkerPk, 0, 1, expectedMinerActorIDAddress)
```
anorth, line 98: The parameter currently `1` depends on how many prior actors have been Exec'd, so I can't quite tell if that needs changing too. Should probably become a parameter to this method or a state variable..

anorth (approved):

drivers/state.go
whyrusleeping: Yeah, probably makes sense. Though at some point we're just reimplementing the whole logic that we're trying to test

PR #132, opened 2020-03-23
Some tipset tests for miner reward and penalties

anorth:


drivers/test.go
```diff
@@ -324,6 +325,11 @@ func (td *TestDriver) applyMessageExpectCodeAndReturn(msg *types.Message, code e
 	return receipt.GasUsed.Int64()
 }
 
+func (td *TestDriver) AssertNoActor(addr address.Address) {
+	_, err := td.State().Actor(addr)
+	require.Error(td.T, err, "expected no such actor %s", addr)
```
frrist, line 335:
```suggestion
	assert.Error(td.T, err, "expected no such actor %s", addr)
```

drivers/test.go
```diff
@@ -435,3 +441,47 @@ func (td *TestDriver) MustCreateAndVerifyMultisigActor(nonce int64, value abi_sp
...
+
+func (r *RewardSummary) For(a address.Address) abi_spec.TokenAmount {
+	v, ok := r.Rewards[a]
+	if !ok {
+		return big_spec.Zero()
+	}
+	return v
+}
+
+func (td *TestDriver) GetRewardSummary() *RewardSummary {
```
frrist, line 472: :star: This is a great idea

suites/tipset/rewards_penalties.go
```diff
@@ -0,0 +1,127 @@
...
+			utils.NewIDAddr(t, 1234),
+			utils.NewSECP256K1Addr(t, "1234"),
+			utils.NewBLSAddr(t, 1234),
+			utils.NewActorAddr(t, "1234"),
+		}
+
+		bb := blkBuilder.WithTicketCount(1)
+		for _, s := range badSenders {
+			bb.WithBLSMessageAndReceipt(td.MessageProducer.Transfer(receiver, s, chain.Value(sendValue)),
+				types.MessageReceipt{ExitCode: exitcode.SysErrActorNotFound, ReturnValue: drivers.EmptyReturnValue},
```
frrist, line 98: I think instead of using `drivers.EmptyReturnValue` we should just pass `nil`. I have a PR that I need to push for removing all usage of `drivers.EmptyReturnValue`, its left over from before we returned `nil`

frrist (approved):

anorth:
Oops, I merged before pushing the suggestions made

PR #135, opened 2020-03-24
implement builder pattern for multiblock tipsets

frrist:


drivers/tipset_message_producer.go
```diff
@@ -11,125 +12,151 @@ import (
...
+	var results []Result
+	for _, b := range t.bbs {
+		results = append(results, b.expectedResults...)
+	}
+
+	if len(receipts) > len(results) {
+		t.driver.T.Fatalf("ApplyTipSetMessages returned more receipts than expected. Expected: %d, Actual: %d", len(results), len(receipts))
+	}
+
+	for i := range receipts {
```
anorth, line 53: It would be a little more helpful to iterate over the blocks, rather than flatten, so that the `(blockIndex, messageIndex)` (before any de-duplication) is available.

suites/tipset/message_dedup.go
```diff
@@ -24,71 +24,81 @@ func TestBlockMessageDeduplication(t *testing.T, factory state.Factories) {
 	t.Run("apply a single BLS message", func(t *testing.T) {
 		td := builder.Build(t)
 		defer td.Complete()
-		blkBuilder := drivers.NewTipSetMessageBuilder(td)
+		tipB := drivers.NewTipSetMessageBuilder(td)
+		blkB := drivers.NewBlockBuilder(td.ExeCtx.Miner)
```
anorth, line 27: It is illegal for a tipset to have more than one block with the same miner, so the miner must be moved out of the test drive execution context, and prepared explicitly.

This change could come in a follow-up.

suites/tipset/message_dedup.go
```diff
@@ -24,71 +24,81 @@ func TestBlockMessageDeduplication(t *testing.T, factory state.Factories) {
...
 		sender, _ := td.NewAccountActor(address.SECP256K1, big_spec.NewInt(10_000_000))
 		receiver, _ := td.NewAccountActor(address.SECP256K1, big_spec.Zero())
 
-		blkBuilder.WithTicketCount(1).
-			// send value from sender to receiver
-			WithBLSMessageOk(
-				td.MessageProducer.Transfer(receiver, sender, chain.Nonce(0), chain.Value(big_spec.NewInt(100))),
-			).ApplyAndValidate()
+		tipB.WithBlockBuilder(
+			blkB.WithTicketCount(1).
```
anorth, line 33: In a follow-up, please set a default of 1 or make it a ctor param, so we can drop this line.

suites/tipset/message_dedup.go
```diff
@@ -24,71 +24,81 @@ func TestBlockMessageDeduplication(t *testing.T, factory state.Factories) {
...
-			// only should have a single result
-			WithResult(exitcode.Ok, drivers.EmptyReturnValue).
-			ApplyAndValidate()
+		tipB.WithBlockBuilder(
+			blkB.WithTicketCount(1).
+				// duplicate the message
+				WithBLSMessageOk(td.MessageProducer.Transfer(receiver, sender, chain.Nonce(0), chain.Value(big_spec.NewInt(100)))).
+				WithBLSMessageOk(td.MessageProducer.Transfer(receiver, sender, chain.Nonce(0), chain.Value(big_spec.NewInt(100)))).
+				// only should have a single result
+				WithResult(exitcode.Ok, drivers.EmptyReturnValue),
```
anorth: Nit: the `WithBLSMessageOk` is confusing here. Is it asserting both the messages are ok or not?

Better would be `WithBLSMessageDropped` or something explicit like that, and then no `WithResult` special handling here.

anorth (approved):

drivers/tipset_message_producer.go
anorth: Another way of saying this: push validation into the BlockBuilder, so that here we iterate over blocks calling ApplyAndValidate, and thus can know the block index and message index in the block. Duplicate messages makes it a little tricky. you might need an explicit null result placeholder for the dropped messages.

frrist: Since this is non-urgent (a possibly holding up other work to avoid rebase-hell) I will add this in a follow on. Nice idea!

PR #137, opened 2020-03-24
Feat/empty val

ZenGround0:
Move expected empty return value to []byte{} from nil


frrist (approved):
LGTM after build passes

PR #138, opened 2020-03-24
Apply Message Returns a Result structure

frrist:
Two things happening here:
1. ApplyMessage returns a result structure containing the message receipt, miner penalty, and gas reward 
2. Message Receipt GasUsed is type GasUnits (int64 alias) with a method `AsBigInt()` since it is almost exclusively compared against bigints.  

In a follow on `ApplyResult` could be extended to contain previous and current state CID's as well as other relevant information.

### DIFF
--- a/chain/types/receipt.go
+++ b/chain/types/receipt.go
@@ -15,6 +15,6 @@ type MessageReceipt struct {
 
 type GasUnits int64
 
-func (gu GasUnits) AsBigInt() big.Int {
+func (gu GasUnits) Big() big.Int {
 	return big.NewInt(int64(gu))
 }
--- a/chain/types/receipt.go
+++ b/chain/types/receipt.go
@@ -10,6 +10,11 @@ type MessageReceipt struct {
 	ExitCode    exitcode.ExitCode
 	ReturnValue []byte
 
-	// TODO make this an int64
-	GasUsed big.Int
+	GasUsed GasUnits
+}
+
+type GasUnits int64
+
+func (gu GasUnits) AsBigInt() big.Int {
+	return big.NewInt(int64(gu))
 }
--- a/chain/validator.go
+++ b/chain/validator.go
@@ -14,8 +14,8 @@ type Validator struct {
 
 type ApplyResult struct {
 	Receipt types.MessageReceipt
-	Penalty state.MinerPenaltyFIL
-	Reward  state.GasRewardFIL
+	Penalty abi.TokenAmount
+	Reward  abi.TokenAmount
 }
 
 // NewValidator builds a new validator.
--- a/chain/validator.go
+++ b/chain/validator.go
@@ -18,7 +18,7 @@ func NewValidator(executor state.Applier) *Validator {
 }
 
 // ApplyMessages applies a message to a state
-func (v *Validator) ApplyMessage(context *types.ExecutionContext, state state.VMWrapper, message *types.Message) (types.MessageReceipt, error) {
+func (v *Validator) ApplyMessage(context *types.ExecutionContext, state state.VMWrapper, message *types.Message) (types.MessageReceipt, state.MinerPenaltyFIL, state.GasRewardFIL, error) {
 	return v.applier.ApplyMessage(context, state, message)
 }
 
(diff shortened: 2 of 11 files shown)


chain/types/receipt.go
```diff
@@ -10,6 +10,11 @@ type MessageReceipt struct {
...
 
-	// TODO make this an int64
-	GasUsed big.Int
+	GasUsed GasUnits
+}
+
+type GasUnits int64
+
+func (gu GasUnits) AsBigInt() big.Int {
+	return big.NewInt(int64(gu))
```
anorth, line 19: `NewIntUnsigned(gu)`

chain/validator.go
```diff
@@ -16,7 +16,7 @@ func NewValidator(executor state.Applier) *Validator {
 }
 
 // ApplyMessages applies a message to a state
-func (v *Validator) ApplyMessage(context *types.ExecutionContext, state state.VMWrapper, message *types.Message) (types.MessageReceipt, error) {
+func (v *Validator) ApplyMessage(context *types.ExecutionContext, state state.VMWrapper, message *types.Message) (types.MessageReceipt, state.MinerPenaltyFIL, state.GasRewardFIL, error) {
```
anorth: You could move ApplyResult here and return it, rather than the four values. I think that would be a little better, tho not dramatic.

state/applier.go
```diff
@@ -9,9 +9,12 @@ import (
 	"github.com/filecoin-project/chain-validation/chain/types"
 )
 
+type MinerPenaltyFIL = abi.TokenAmount
+type GasRewardFIL = abi.TokenAmount
```
anorth: I don't think these are necessary. As aliases, there's no actual safety, and they are the same logical type (tokens).

If you return a structure from ApplyMessage, I don't think any justification would remain.

suites/message/create_actor.go
```diff
@@ -86,15 +86,15 @@ func TestAccountActorCreation(t *testing.T, factory state.Factories) {
...
+			result := td.ApplyFailure(
 				td.MessageProducer.Transfer(tc.newActorAddr, existingAccountAddr, chain.Value(tc.newActorInitBal), chain.Nonce(0)),
 				tc.expExitCode,
 			)
 
 			// new actor balance will only exist if message was applied successfully.
 			if tc.expExitCode.IsSuccess() {
 				td.AssertBalance(tc.newActorAddr, tc.newActorInitBal)
-				td.AssertBalance(existingAccountAddr, big_spec.Sub(big_spec.Sub(tc.existingActorBal, big_spec.NewInt(gasUsed)), tc.newActorInitBal))
+				td.AssertBalance(existingAccountAddr, big_spec.Sub(big_spec.Sub(tc.existingActorBal, result.Receipt.GasUsed.AsBigInt()), tc.newActorInitBal))
```
anorth: This is fine for now. An alternative would be `GasUsed.Cost(gasPrice)`, which would make this a little more obvious to the uninitiated.

chain/types/receipt.go
```diff
@@ -10,6 +10,11 @@ type MessageReceipt struct {
...
 	ReturnValue []byte
 
-	// TODO make this an int64
-	GasUsed big.Int
+	GasUsed GasUnits
+}
+
+type GasUnits int64
+
+func (gu GasUnits) AsBigInt() big.Int {
```
anorth: How about `AsBig()`, or even just `Big()`?

anorth (approved):

chain/types/receipt.go
frrist: GasUnits are signed, right?

suites/message/create_actor.go
frrist: Will push to a follow on

state/applier.go
frrist: Agreed, removed.

chain/types/receipt.go
anorth: Oh, yes sorry. I'm not clear why you need the conversion in this case then.

frrist: because of `type GasUnits int64` on line 16

PR #139, opened 2020-03-24
update gas values

frrist:
- gas values were generated from go-filecoin after fixing bls seed bug. related PR: https://github.com/filecoin-project/go-filecoin/pull/3923


anorth (approved):

box/blob.go
```diff
@@ -16,10 +16,13 @@ func init() {
 	resources.Add("/TestChainValidationMessageSuiteTestMultiSigActorconstructortest", []byte{49, 56, 53, 49, 10})
 	resources.Add("/TestChainValidationMessageSuiteTestMultiSigActorproposeandapprove", []byte{50, 48, 51, 55, 10, 56, 57, 53, 10, 50, 50, 50, 10, 49, 54, 56, 10, 49, 48, 50, 55, 10})
 	resources.Add("/TestChainValidationMessageSuiteTestMultiSigActorproposeandcancel", []byte{50, 48, 51, 55, 10, 56, 57, 53, 10, 50, 50, 50, 10, 52, 51, 56, 10})
-	resources.Add("/TestChainValidationMessageSuiteTestNestedSendsokbasicsend", []byte{})
 	resources.Add("/TestChainValidationMessageSuiteTestPaychhappypathcollect", []byte{49, 57, 49, 56, 10, 51, 50, 48, 10, 49, 57, 49, 10, 50, 51, 54, 10})
 	resources.Add("/TestChainValidationMessageSuiteTestPaychhappypathconstructor", []byte{49, 57, 49, 56, 10})
 	resources.Add("/TestChainValidationMessageSuiteTestPaychhappypathupdate", []byte{49, 57, 49, 56, 10, 51, 49, 56, 10})
+	resources.Add("/TestChainValidationMessageSuiteTestValueTransferAdvancefailtotransferfromunknownaccounttoknownaddress", []byte{48, 10})
```
anorth, line 22: Are these bytes supposed to represent integers? Using BigInt (maybe parsed from string) would be easier to understand.

frrist: They represent the recorded files as bytes (so yes, currently they are integers). Leaving the implementation like this will give us greater flexibility to change the contents of those files later. (e.g. we could add state roots in addition to gas values). I'd prefer to leave this as it exists now.

PR #141, opened 2020-03-25
Update specs-actors with new miner-penalty exit code, less use of EmptyValue

anorth:
Significant changes are:
- use `nil` instead of instances of EmptyValue
- new exit codes for miner-penalty send failures
- regenerated gas values (because the nil parameter is cheaper than EmptyValue)

I checked in the intermediate gas files as @frrist had noted they'd be handy while debugging (the box generated file isn't human-readable).

FYI @whyrusleeping


box/gen.go
```diff
@@ -90,5 +90,4 @@ func main() {
 	}
 
 	log.Println("Baking resources done... \U0001F680")
-	log.Println("DO NOT COMMIT box/blob.go \U0001F47B")
```
anorth: @frrist I don't know what this scary-looking line is for. The file is already committed.

chain/account.go
```diff
@@ -12,7 +12,7 @@ func (mp *MessageProducer) AccountConstructor(to, from address.Address, params a
 	ser := MustSerialize(&params)
 	return mp.Build(to, from, builtin_spec.MethodsAccount.Constructor, ser, opts...)
 }
-func (mp *MessageProducer) AccountPubkeyAddress(to, from address.Address, params adt.EmptyValue, opts ...MsgOpt) *types.Message {
-	ser := MustSerialize(&params)
+func (mp *MessageProducer) AccountPubkeyAddress(to, from address.Address, params *adt.EmptyValue, opts ...MsgOpt) *types.Message {
```
anorth, line 15: So I can pass `nil`. I do think all the others should change to be pointers too, but not urgent enough to do now.

box/gen.go
frrist: Its not important, removal here is file. The scary bit at the end is an emoji.

box/resources/TestChainValidationMessageSuiteTestAccountActorCreationfailcreateBLSaccountactorinsufficientbalance
```diff
@@ -0,0 +1 @@
+0
```
frrist: ask discussed afk, once this is rebased on master these files can all be removed since blob.go is human readable in latest change

frrist (approved):

box/resources/TestChainValidationMessageSuiteTestAccountActorCreationfailcreateBLSaccountactorinsufficientbalance
anorth: Done

arajasek (approved):
LGTM

PR #145, opened 2020-03-25
add message application methods for signed messsages

frrist:


drivers/test.go
```diff
@@ -330,6 +335,69 @@ func (td *TestDriver) applyMessageExpectCodeAndReturn(msg *types.Message, code e
...
+	}
+	if td.Config.ValidateReturnValue() {
+		assert.Equal(td.T, retval, result.Receipt.ReturnValue, "Expected ReturnValue: %v Actual ReturnValue: %v", retval, result.Receipt.ReturnValue)
+	}
+	if td.Config.ValidateGas() {
+		expectedGasUsed, ok := td.GasMeter.NextExpectedGas()
+		if ok {
+			assert.Equal(td.T, expectedGasUsed, result.Receipt.GasUsed, "Expected GasUsed: %d Actual GasUsed: %d", expectedGasUsed, result.Receipt.GasUsed)
+		} else {
+			td.T.Logf("WARNING: failed to find expected gas cost for message: %+v", msg)
```
frrist: This code is duplicated because we need need `msg` for a helpful debug message, could remove?

chain/types/message.go
```diff
@@ -21,18 +21,18 @@ type Message struct {
...
 	// Optional method to invoke on receiver, zero for a plain value send.
 	Method abi.MethodNum
 	/// Serialized parameters to the method (if method is non-zero).
 	Params []byte
-
-	GasPrice big.Int
-	GasLimit int64
 }
 
 func (t *Message) MarshalCBOR(w io.Writer) error {
```
frrist, line 41: I copied these marshalers from lotus as it was easier than writing the code gen here.

anorth: Please write something like this in the code, not just this PR.

drivers/test.go
anorth: Yea prefer to de-dupe. You can use an `interface{}` for the message debugging?

anorth (approved):

drivers/test.go
frrist: how about returning a boolean value, false if gas isn't found?

arajasek:
LGTM

arajasek (approved):

PR #146, opened 2020-03-25
block builder internal message signing

frrist:


drivers/tipset_message_producer.go
```diff
@@ -141,23 +133,55 @@ func (bb *BlockBuilder) WithBLSMessageAndRet(bm *types.Message, retval []byte) *
...
+
 func (bb *BlockBuilder) WithTicketCount(count int64) *BlockBuilder {
 	bb.ticketCount = count
 	return bb
 }
 
+func (bb *BlockBuilder) toSignedMessage(m *types.Message) *types.SignedMessage {
+	from := m.From
+	if from.Protocol() != address.SECP256K1 {
+		bb.TD.T.Fatalf("message From field must be SECP256K1 protocol, was: %v", from.Protocol())
```
anorth: Sorry, this restriction prevents us testing important valid cases when `From` is an ID address of an account actor. We need to be able to resolve the ID address to corresponding account.

drivers/tipset_message_producer.go
```diff
@@ -90,8 +92,9 @@ type Result struct {
 	ReturnVal []byte
 }
 
-func NewBlockBuilder(miner address.Address) *BlockBuilder {
+func NewBlockBuilder(miner address.Address, td *TestDriver) *BlockBuilder {
```
anorth: Swap parameter order

anorth (changes requested):

drivers/tipset_message_producer.go
frrist: Would you expect that lookup be performed by inspecting the init actors state or should chain-validation keep its own mapping?

anorth:
Probably the latter

drivers/tipset_message_producer.go
frrist: Addressed in latest set of changes.

drivers/state.go
```diff
@@ -32,11 +32,14 @@ type StateDriver struct {
 	st state.VMWrapper
 	w  state.KeyManager
 	rs state.RandomnessSource
+
+	// Mapping for IDAddresses to their pubkey/actor addresses. Used for lookup when signing messages.
+	ActorIDMap map[address.Address]address.Address
```
anorth: Don't export this. Provide a function if lookup from outside is needed.

anorth (approved):

PR #148, opened 2020-03-26
More inner send failure tests

anorth:
~Note: go-filecoin doesn't pass these yet so I don't have gas values or 100% certainty of correctness.~

### DIFF
--- a/drivers/test.go
+++ b/drivers/test.go
@@ -427,9 +427,8 @@ func (td *TestDriver) AssertActor(addr address.Address, balance abi_spec.TokenAm
 }
 
 func (td *TestDriver) AssertHead(addr address.Address, expected cid.Cid) {
-	actr, err := td.State().Actor(addr)
-	require.NoError(td.T, err)
-	assert.Equal(td.T, expected, actr.Head(), "expected actor %s head %s, actual %s", addr, expected, actr.Head())
+	head := td.GetHead(addr)
+	assert.Equal(td.T, expected, head, "expected actor %s head %s, actual %s", addr, expected, head)
 }
 
 func (td *TestDriver) AssertBalanceCallback(addr address.Address, thing func(actorBalance abi_spec.TokenAmount) bool) {
--- a/drivers/test.go
+++ b/drivers/test.go
@@ -405,6 +405,12 @@ func (td *TestDriver) GetBalance(addr address.Address) abi_spec.TokenAmount {
 	return actr.Balance()
 }
 
+func (td *TestDriver) GetHead(addr address.Address) cid.Cid {
+	actr, err := td.State().Actor(addr)
+	require.NoError(td.T, err)
+	return actr.Head()
+}
+
 // AssertBalance checks an actor has an expected balance.
 func (td *TestDriver) AssertBalance(addr address.Address, expected abi_spec.TokenAmount) {
 	actr, err := td.State().Actor(addr)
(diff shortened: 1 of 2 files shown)


drivers/test.go
```diff
@@ -356,6 +362,12 @@ func (td *TestDriver) AssertActor(addr address.Address, balance abi_spec.TokenAm
 	assert.Equal(td.T, callSeqNum, actr.CallSeqNum(), fmt.Sprintf("expected actor %s callSeqNum: %d, actual : %d", addr, callSeqNum, actr.CallSeqNum()))
 }
 
+func (td *TestDriver) AssertHead(addr address.Address, expected cid.Cid) {
+	actr, err := td.State().Actor(addr)
+	require.NoError(td.T, err)
```
frrist: nit: I think you could reuse `GetHead` here.

suites/message/nested.go
```diff
@@ -210,23 +212,126 @@ func TestNestedSends(t *testing.T, factory state.Factories) {
 		td.AssertBalance(stage.creator, big.Sub(balanceBefore, result.Receipt.GasUsed.Big())) // Pay gas, don't receive funds.
 	})
 
+	// The multisig actor checks before attempting to transfer more than its balance, so we can't exercise that
+	// the VM also checks this. Need a custome actor to exercise this.
```
frrist, line 217: Clarifying the thing we want to exercise is that inner sends for an amount greater than actor balance fail as expected, correct? In a similar vein to this, do we exercise inner sends of adequate balance for transfer but inadequate to cover gas cost?

suites/message/nested.go
```diff
@@ -210,23 +212,126 @@ func TestNestedSends(t *testing.T, factory state.Factories) {
...
+	t.Run("fail missing params", func(t *testing.T) {
+		td := builder.Build(t)
+		defer td.Complete()
+
+		stage := prepareStage(td, acctDefaultBalance, multisigBalance)
+		balanceBefore := td.GetBalance(stage.creator)
+
+		params := adt.Empty // Missing params required by AddSigner
+		amtSent := abi.NewTokenAmount(1)
+		result := stage.send(stage.msAddr, amtSent, builtin.MethodsMultisig.AddSigner, params, 1)
```
frrist: should this be `&params`?

suites/message/nested.go
```diff
@@ -210,23 +212,126 @@ func TestNestedSends(t *testing.T, factory state.Factories) {
...
+			From: builtin.SystemActorAddr,
+			To:   builtin.SystemActorAddr,
+		}
+		execParams := init_.ExecParams{
+			CodeCID:           builtin.PaymentChannelActorCodeID,
+			ConstructorParams: chain.MustSerialize(&ctorParams),
+		}
+
+		amtSent := abi.NewTokenAmount(1)
+		result := stage.send(builtin.InitActorAddr, amtSent, builtin.MethodsInit.Exec, &execParams, 1)
```
frrist: nit: `approverNonce` is constant throughout these tests (`1`), could they instead use a variable that indicates what `1` is?

frrist (commented):
Looking good! Will give another look when out of draft.

arajasek:
FYI Lotus passes all of these tests, so that's some measure of confidence in correctness. We'll see how gas values go.

suites/message/nested.go
anorth: the inner sender doesn't pay for gas.

suites/message/nested.go
anorth: Not, adt.Empty is a nil pointer

frrist (approved):

PR #149, opened 2020-03-26
add state root tracking to chain-val

frrist:


anorth:
Gas and state roots should be tracked together, in the same file and at the same intervals. I don't think the duplication of the box mechanism and all the related stuff is ok.

It looks like you might need a refactor to the tipset case to enable this, propagating the message `Result`s rather than only receipts. I expect that to be much less code than this introduces, though.

frrist:
replaced by https://github.com/filecoin-project/chain-validation/pull/150

PR #150, opened 2020-03-27
add state root tracking and refactor gasmeter -> state tracker

frrist:


box/box.go
```diff
@@ -5,11 +5,11 @@
 package box
 
 type resourceBox struct {
```
frrist, line 7: resource box is now generic store, it contains `ApplyMessageResult`'s and `ApplyTipSetMessagesResult`s. When we want to add another value to track we can put it in one of the result structures.

drivers/test.go
```diff
@@ -344,45 +344,53 @@ func (td *TestDriver) ApplyMessageSigned(msg *types.Message) (result chain.Apply
...
-		expectedGasUsed, ok := td.GasMeter.NextExpectedGas()
+		expectedGasUsed, ok := td.StateTracker.NextExpectedGas()
 		if ok {
 			assert.Equal(td.T, expectedGasUsed, result.Receipt.GasUsed, "Expected GasUsed: %d Actual GasUsed: %d", expectedGasUsed, result.Receipt.GasUsed)
 		} else {
 			foundGas = false
 		}
 	}
-	// TODO in the very near future we will be validating the stateroot here, keep in back of head.
+	if td.Config.ValidateStateRoot() {
```
frrist, line 385: The only meaningful change in this file to wire up stateroot validation.

tracker/track.go
```diff
@@ -0,0 +1,175 @@
...
+	switch v := data.(type) {
+	case chain.ApplyMessageResult:
+		root, err := cid.Decode(v.Root)
+		if err != nil {
+			t.Fatal(err)
+		}
+		gasUsed = append(gasUsed, v.Receipt.GasUsed)
+		stateRoots = append(stateRoots, root)
+		return
+	case []chain.ApplyMessageResult:
```
anorth, line 114: The slice/single is annoying. Is there any way to get it to always be a slice?

chain/validator.go
```diff
@@ -12,10 +12,16 @@ type Validator struct {
...
 
-type ApplyResult struct {
+type ApplyMessageResult struct {
 	Receipt types.MessageReceipt
 	Penalty abi.TokenAmount
 	Reward  abi.TokenAmount
+	Root    string
+}
+
+type ApplyTipSetMessagesResult struct {
```
anorth: ApplyTipSetResult is probably enough, and briefer.

anorth (approved):

tracker/track.go
frrist: I suppose for the case that it _is_ a single could just shove it in a slice and reuse the same logic.

arajasek (approved):
Yup, looks good to me! Thanks!

PR #152, opened 2020-03-31
Update state validations for gas payment staging with reward rather than burnt funds actor

anorth:
See corresponding change in go-filecoin: https://github.com/filecoin-project/go-filecoin/pull/3948

Lotus passes all the message validation tests (incl state root CID comparison) after this change.


frrist (approved):
lgtm

PR #154, opened 2020-04-07
refactor Applier interface and consolidate types

frrist:
This PR refactors the Applier interface with the following changes:
- Applying a message/signed message returns an `ApplyMessageResult`
- Applying a tipset returns an `ApplyTipSetResult`
- The `Apply*Result` types now live in `chain/types/`
- `ApplyMessage` and `ApplySignedMessage` now accept just a epoch and a message to apply.
  - To accommodate this change implementations need to create their applier and state driver together.
- `Factories` has a single method to create an applier and vmwrapper: `NewStateAndApplier`.
- `RandomnessSource` has been removed from the `Factories` interface as it was unused.

I have 2 follow ons for this in lotus and go-filecoin that adapt their drivers to work with these changes (tests pass). I will push them up if once this goes in.


tracker/track.go
```diff
@@ -12,7 +12,6 @@ import (
 	"github.com/ipfs/go-cid"
```
frrist, line 12: `chain` -> `types`

suites/message/nested.go
```diff
@@ -21,6 +21,7 @@ import (
 	typegen "github.com/whyrusleeping/cbor-gen"
```
frrist, line 21: `chain` -> `types`

box/blob.go
```diff
@@ -5,57 +5,56 @@ package box
 import (
 	"github.com/filecoin-project/specs-actors/actors/abi"
 
-	"github.com/filecoin-project/chain-validation/chain"
```
frrist, line 8: `chain` -> `types`

state/factories.go
```diff
@@ -2,13 +2,9 @@ package state
 
 // Factories wraps up all the implementation-specific integration points.
 type Factories interface {
-	NewState() VMWrapper
+	NewStateAndApplier() (VMWrapper, Applier)
```
frrist, line 5: Previously, methods on `Applier` accepted a `VMWrapper`. Now an `Applier` contains a `VMWrapper` and these need to be created at the same time, so this methods was added.

PR #168, opened 2020-04-14
[wip] partial update to work with new specs-actors

frrist:
- There are two FIXME's that need to be addressed before merging
  - drivers/test.go: GetRewardSummary()
  - suites/tipset/miner_flow.go: TestMinerSubmitFallbackPoSt()

### DIFF
--- a/drivers/test.go
+++ b/drivers/test.go
@@ -28,7 +28,6 @@ import (
 	cbor "github.com/ipfs/go-ipld-cbor"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	cbg "github.com/whyrusleeping/cbor-gen"
 
 	"github.com/filecoin-project/chain-validation/chain"
 	"github.com/filecoin-project/chain-validation/chain/types"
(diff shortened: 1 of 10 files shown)


suites/tipset/miner_flow.go
```diff
@@ -272,5 +268,6 @@ func TestMinerSubmitFallbackPoSt(t *testing.T, factory state.Factories) {
 
 	var powerSt power_spec.State
 	td.GetActorState(builtin_spec.StoragePowerActorAddr, &powerSt)
-	assert.Equal(t, drivers.InitialTotalNetworkPower+int64(32<<30), powerSt.TotalNetworkPower.Int64())
+	// FIXME: ensure this is the correct assertion to make after the changes that landed in specs-actors
```
frrist, line 271: need to address before merging

drivers/test.go
```diff
@@ -550,31 +557,36 @@ func (r *RewardSummary) For(a address.Address) abi_spec.TokenAmount {
...
-	err := adt_spec.AsMap(AsStore(td.State()), rst.RewardMap).ForEach(&r, func(key string) error {
-		keyAddr, err := address.NewFromBytes([]byte(key))
-		require.NoError(td.T, err)
-
-		// Traverse array of reward entries.
-		sum := big_spec.Zero()
-		var rw reward_spec.Reward
-		err = adt_spec.AsArray(AsStore(td.State()), cid.Cid(r)).ForEach(&rw, func(i int64) error {
-			sum = big_spec.Sub(big_spec.Add(sum, rw.Value), rw.AmountWithdrawn)
+	// FIXME the reward actor's responsibilities were stripped down, this method needs to be adjusted to reflect that
```
frrist, line 560: need to fix before merging

anorth: The rewards earned by a miner have moved to the miner actor as `LockedFunds`. I think inspecting that was the primary use of this structure.

We might still want some summary of the reward actor's new state, but we can start without it.

suites/tipset/miner_flow.go
anorth: You want to inspect the TotalRawBytePower in power state, but then I think it should be still good.

frrist:
closing in favor of #170

PR #170, opened 2020-04-22
update specs actors v1

frrist:


suites/tipset/rewards_penalties.go
```diff
@@ -67,7 +69,7 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
 				// No reward is paid to the miner directly. The funds for block reward were already held by the
 				// reward actor. The gas reward should be added to the treasury. The sum of block and gas reward
 				// should be chalked up to the miner address.
-				thisReward := big.Add(reward.BlockRewardTarget, gasSum)
+				thisReward := big.Add(BlockRewardTarget, gasSum)
```
frrist: @anorth was `reward.BlockRewardTarget` the value of the block reward or something different? What do we call it now if not this?

drivers/test.go
```diff
@@ -530,11 +540,12 @@ func (td *TestDriver) MustCreateAndVerifyMultisigActor(nonce uint64, value abi_s
 		UnlockDuration:        params.UnlockDuration,
 		NumApprovalsThreshold: params.NumApprovalsThreshold,
 
-		PendingTxns: pendingTxMap.Root(),
+		PendingTxns: pendingTxMapRoot,
 	})
 	td.AssertBalance(multisigAddr, value)
 }
 
+// XXX: changes needed here
```
frrist: @anorth based on the above tests I believe a structure like this is still useful. My guess is that `Treasury` and `RewardsTotal` are not longer required?

Since BlockRewards are now paid directly to miners (instead of being brokered via the reward actor) I think the `Rewards` map could be populated by inspecting each miners vesting store at a specific epoch (I may be way of here)?

chain/miner.go
```diff
@@ -15,8 +15,8 @@ func (mp *MessageProducer) MinerConstructor(to, from address.Address, params pow
 	ser := MustSerialize(&params)
 	return mp.Build(to, from, builtin_spec.MethodsMiner.Constructor, ser, opts...)
 }
-func (mp *MessageProducer) MinerControlAddresses(to, from address.Address, params *adt.EmptyValue, opts ...MsgOpt) *types.Message {
-	ser := MustSerialize(params)
+func (mp *MessageProducer) MinerControlAddresses(to, from address.Address, params adt.EmptyValue, opts ...MsgOpt) *types.Message {
```
anorth: IMO this is a regression. _All_ the params objects should be passed by pointer, we just didn't fix the rest yet.

drivers/state.go
```diff
@@ -124,52 +124,63 @@ func (d *StateDriver) newMinerAccountActor() address.Address {
...
+		VestingFunds:        EmptyArrayCid,
+		Sectors:             EmptyArrayCid,
+		SectorExpirations:   EmptyArrayCid,
+		FaultEpochs:         EmptyArrayCid,
+		Deadlines:           EmptyDeadlinesCid,
+
+		NewSectors:      abi_spec.NewBitField(),
+		Faults:          abi_spec.NewBitField(),
+		Recoveries:      abi_spec.NewBitField(),
+		PostSubmissions: abi_spec.NewBitField(),
```
anorth: Using `ConstructState` would be much more robust. Is that possible?

drivers/state.go
```diff
@@ -124,52 +124,63 @@ func (d *StateDriver) newMinerAccountActor() address.Address {
...
-		FaultSet:            abi_spec.NewBitField(),
-		ProvingSet:          EmptyArrayCid,
 		Info: miner_spec.MinerInfo{
 			Owner:            minerOwnerID,
 			Worker:           minerWorkerID,
 			PendingWorkerKey: nil,
 			PeerId:           "chain-validation",
 			SectorSize:       0,
+			// XXX: verify
+			ProvingPeriodBoundary: 0,
```
anorth: This is _a_ valid value, but should probably be a parameter to this method.

drivers/state.go
```diff
@@ -124,52 +124,63 @@ func (d *StateDriver) newMinerAccountActor() address.Address {
...
-		PreCommittedSectors: EmptyMapCid,
-		Sectors:             EmptyArrayCid,
-		FaultSet:            abi_spec.NewBitField(),
-		ProvingSet:          EmptyArrayCid,
 		Info: miner_spec.MinerInfo{
 			Owner:            minerOwnerID,
 			Worker:           minerWorkerID,
 			PendingWorkerKey: nil,
 			PeerId:           "chain-validation",
 			SectorSize:       0,
```
anorth: This value isn't valid. I suggest using 2048 if you need a constant (define this in a constant somewhere)

drivers/test.go
```diff
@@ -91,10 +92,10 @@ func init() {
 		Balance: big_spec.Zero(),
 		Code:    builtin_spec.StoragePowerActorCodeID,
 		State: &power_spec.State{
-			TotalNetworkPower:        abi_spec.NewStoragePower(InitialTotalNetworkPower),
-			EscrowTable:              EmptyMapCid,
+			TotalRawBytePower:        abi_spec.NewStoragePower(InitialTotalNetworkPower),
```
anorth: Can you please file an issue for whatever the reason is that this can't be zero? Where's the panic?

drivers/test.go
anorth: Treasury is still relevant and easy to get. Drop the reward and total and replace with SimpleSupply, BaselineSupply, LastPerEpochReward from reward actor state.

Change call sites that used this to find the miner's reward to get the miner actor's balance directly, via existing functions.

suites/tipset/miner_flow.go
```diff
@@ -171,7 +172,8 @@ func TestMinerMissPoStChallengeWindow(t *testing.T, factory state.Factories) {
 	}
 
 	// Epoch advances to the end of the proving window. Send a sing message to trigger the cron actor send
-	td.ExeCtx.Epoch += power_spec.WindowedPostChallengeDuration + miner_spec.ProvingPeriod
+	// XXX: verify this is correct
+	td.ExeCtx.Epoch += miner_spec.WPoStChallengeWindow + 1
```
anorth: Do you want to test missing a single PoSt deadline, or missing all of them for a whole period? The above will do the latter.

The right epoch depends on the miner's proving period boundary too.

I'm guessing this won't automatically invoke cron. You need to have the power actor's cron callback invoked to trigger miner cleanup. The code might assume that it's invoked for _every_ epoch, no gaps.

suites/tipset/miner_flow.go
```diff
@@ -185,8 +187,12 @@ func TestMinerMissPoStChallengeWindow(t *testing.T, factory state.Factories) {
...
 	var minerSt miner_spec.State
 	td.GetActorState(minerActorID, &minerSt)
-	require.True(t, minerSt.PoStState.HasFailedPost())
-	require.Equal(t, int64(1), minerSt.PoStState.NumConsecutiveFailures)
+	// TODO refigure this
+	/*
+		require.True(t, minerSt.PoStState.HasFailedPost())
+		require.Equal(t, int64(1), minerSt.PoStState.NumConsecutiveFailures)
+
+	*/
```
anorth: Sectors should all be marked faulty, and miner should have paid fees. Miner should have no claimed power.

suites/tipset/miner_flow.go
```diff
@@ -234,9 +240,11 @@ func TestMinerSubmitFallbackPoSt(t *testing.T, factory state.Factories) {
 		return ownerCallSeq
 	}
 
-	candidates := []abi_spec.PoStCandidate{{
+	candidates := []abi_spec.SectorInfo{{
 		RegisteredProof: abi_spec.RegisteredProof_StackedDRG32GiBPoSt,
-		ChallengeIndex:  0,
+		SectorNumber:    abi_spec.SectorNumber(0),
+		// XXX: probably wrong
+		SealedCID: cid.Undef,
```
anorth: This should come from the proven committed sector created above.

This method generally is all wrong now. You need to work out which deadline the sector is due at and post the proof at the right time. (Don't hardcode any numbers here, the sector->deadline allocation  will change).

suites/tipset/rewards_penalties.go
anorth: It was. You can now get the value of the prior block reward from the reward actor state.

The comment above here is obviously all wrong now: the reward is paid to the miner.

chain/miner.go
frrist: Alright. As a final commit to this PR I will regenerate all these and with params passes by reference.

suites/tipset/miner_flow.go
frrist: The invocation of the cron actor happens below on line 177 (after incrementing the epoch here) by sending a message.

I think a refactor of this test is inorder, I want to do the simplest thing here that allows the code to compile and adds value to impls using this library.

drivers/test.go
frrist: I think it can happen here: https://github.com/filecoin-project/specs-actors/blob/master/actors/builtin/power/policy.go#L52

After this PR is compilable and works against an impl I will file an issue with details if the problem persists.

box/blob.go
```diff
@@ -9,52 +9,50 @@ import (
...
+	resources.Add("/TestChainValidationTipSetSuiteTestMinerRewardsAndPenaltiespenalizesendernonaccount", []types.ApplyTipSetResult{types.ApplyTipSetResult{Receipts: []types.MessageReceipt{types.MessageReceipt{ExitCode: 1, ReturnValue: []uint8{}, GasUsed: 0}, types.MessageReceipt{ExitCode: 1, ReturnValue: []uint8{}, GasUsed: 0}, types.MessageReceipt{ExitCode: 1, ReturnValue: []uint8{}, GasUsed: 0}}, Root: "bafy2bzaced2k5iddb27idt6dhisv2df4pvgzph5q2o5um344yr7ksbgc4tj5g"}})
+	resources.Add("/TestChainValidationTipSetSuiteTestMinerRewardsAndPenaltiespenalizewrongcallseqnum", []types.ApplyTipSetResult{types.ApplyTipSetResult{Receipts: []types.MessageReceipt{types.MessageReceipt{ExitCode: 2, ReturnValue: []uint8{}, GasUsed: 0}}, Root: "bafy2bzacea2cecykfvghcxhggmhhgmvl2g32mlmuxt4umgzogxobbknqfk3bq"}})
```
frrist: the removed miner tests are omitted from this list

drivers/state.go
```diff
@@ -117,59 +117,46 @@ func (d *StateDriver) ActorPubKey(idAddress address.Address) address.Address {
...
-			PeerId:           "chain-validation",
-			SectorSize:       0,
-		},
-		PoStState: miner_spec.PoStState{
-			ProvingPeriodStart:     -1,
-			NumConsecutiveFailures: 0,
-		},
-	})
+	// create the miner actor s.t. it exists in the init actors map
+	minerState := miner_spec.ConstructState(EmptyArrayCid, EmptyMapCid, EmptyDeadlinesCid, minerOwnerID, minerWorkerID, "chain-validation", sectorSize, periodBoundary)
```
frrist, line 128: using the constructor here now as requested.

drivers/test.go
frrist: Confirmed this no longer causes a panic, value is now set to `0`.

suites/tipset/miner_flow.go
```diff
@@ -1,276 +0,0 @@
-package tipset
```
frrist, line 1: Will add a test similar to this in a follow on. Given the amount that changed in the miner actor, and the amount of :red_circle: :green_circle: in this PR I think this is reasonable.

drivers/test.go
```diff
@@ -57,7 +58,9 @@ var (
 )
 
 // if this number is 0 we get a specs-actors panic since it divides by 0
```
anorth: Delete dangling comment.

anorth (approved):

PR #176, opened 2020-04-28
test insufficient gas to cover return value

frrist:


suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,39 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
+		// message and operation cost is 458 gas units
+		// return value costs 56 units of gas, the above isn't enough to cover return value.
+		retval := chain.MustSerialize(&miner_spec.GetControlAddressesReturn{
+			Owner:  td.StateDriver.BuiltinMinerInfo().OwnerID,
+			Worker: td.StateDriver.BuiltinMinerInfo().WorkerID,
+		})
+		bb.WithBLSMessageAndCodeAndReturn(
+			td.MessageProducer.MinerControlAddresses(miner, alice,
+				nil, chain.Nonce(0), chain.GasPrice(gasPrice), chain.GasLimit(gasLimit)),
+			exitcode.SysErrOutOfGas,
```
frrist: Flagging for discussion: We cannot assert the error code `SysErrOutOfGas` was returned due to lack of gas for covering the return value. This makes the test somewhat unreliable. Thoughts?

anorth: Yes. We don't need to hardcode a value here. Send the message once with a high gas limit and record what the actual gas cost was. Then send it again with a gas limit one lower than that.

It wouldn't test the right thing if the return value was free, but I think good enough.

anorth (changes requested):

box/blob.go
```diff
@@ -50,11 +50,9 @@ func init() {
...
...0x67, 0x83, 0x6, 0x36, 0xb4, 0xd2, 0x4f, 0xac, 0xe1}, GasUsed: 3130}, types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{}, GasUsed: 1202}, types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{}, GasUsed: 1153}}, Root: "bafy2bzacedcvzxvecicliig7ddndj3krqnnklmobeks7ktar35dcq7xqgcafi"}, types.ApplyTipSetResult{Receipts: []types.MessageReceipt{types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{0x81, 0x81, 0x0}, GasUsed: 3104}, types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{}, GasUsed: 1984}}, Root: "bafy2bzacechxeeosjhxw6k75ctzytyikzmcz3prnijjbeltt275o6li2j32gu"}, types.ApplyTipSetResult{Receipts: []types.MessageReceipt{types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{}, GasUsed: 8337}}, Root: "bafy2bzacec74uztwvseanh4sctolzcnioophsoiljhh3sp2hphcza2qwwk6ke"}, types.ApplyTipSetResult{Receipts: []types.MessageReceipt{types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{}, GasUsed: 2249}}, Root: "bafy2bzacecdlosbfs3zji4osefjtydsdbxj5x6ob532azt3j3xgxc6chhrkha"}, types.ApplyTipSetResult{Receipts: []types.MessageReceipt{types.MessageReceipt{ExitCode: 0, ReturnValue: []uint8{}, GasUsed: 124}}, Root: "bafy2bzaced2se65zgchy6qj4pcm7rcl2c7w3nbqddc4gxer6j7uqyo6mlax5c"}})
```
frrist: These tests no longer exist so removing them

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,56 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
 		td.AssertBalance(builtin.BurntFundsActorAddr, big.Add(halfBalance, gasPenalty))
 	})
 
+	t.Run("insufficient gas to cover return value", func(t *testing.T) {
```
frrist, line 212: While writing this test the following bug was found in lotus: https://github.com/filecoin-project/lotus/pull/1638. Unfortunately even with the aforementioned fix lotus doesn't pass this test, and go-filecoin doesn't compile against this version of chain-validation, so I cannot generate the resource files for this.

I need to dig in more and see why lotus fails, from the initial debugging I have done it appears the miner is being paid 522 more FIL than it should, this is the value of [gasUnitsUsed](https://github.com/filecoin-project/chain-validation/pull/176/files#diff-01cdcbcb836fde88bea237f4d0cc0364R246)

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,61 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
```
frrist, line 274: blocked on https://github.com/filecoin-project/lotus/pull/1638 to generate resource files for this test.

anorth: I think you linked to the wrong PR. Generally I infer that you're blocked on generating gas values from _some_ implementation, and I'll try to do so for you from go-filecoin.

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,61 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
+		})
+
+		prevRewards1 := td.GetRewardSummary()
+
+		// get a successful result so we can determine how much gas it costs. We'll reduce this by 1 in a subsequent call
+		// to test insufficient gas to cover return value.
+		result1 := tb.WithBlockBuilder(
+			drivers.NewBlockBuilder(td, td.ExeCtx.Miner).
+				WithBLSMessageAndRet(td.MessageProducer.MinerControlAddresses(miner, alice, nil, chain.Nonce(0)),
+					retval,
```
anorth: Related to discussion on #180, validating the return value is orthogonal to the point of this test, and adds noise. It would be better to just call `Apply()`, check the exit code, and ignore the return value itself.

However from a quick scan it looks like the TipSetMessageBuilder has exactly the same error addressed in #180.

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,61 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
+				),
+		).ApplyAndValidate()
+		newRewards1 := td.GetRewardSummary()
+		validateRewards(td, prevRewards1, newRewards1, miner, result1.Receipts[0].GasUsed.Big(), big.Zero())
+		td.AssertBalance(builtin.BurntFundsActorAddr, big.Zero())
```
anorth: These validations are also unnecessary.

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,61 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
+				WithBLSMessageAndCode(td.MessageProducer.MinerControlAddresses(miner, alice,
+					nil, chain.Nonce(1), chain.GasPrice(1), chain.GasLimit(int64(gasUnitsUsed)-1)),
+					exitcode.SysErrOutOfGas,
+				),
+		).ApplyAndValidate()
+
+		newRewards2 := td.GetRewardSummary()
+
+		penalty := big.Zero()
+		rwd := big.Add(big.Sub(newRewards2.LastPerEpochReward, penalty), big.Add(result1.Receipts[0].GasUsed.Big(), result2.Receipts[0].GasUsed.Big()))
```
anorth: This is confusing. Why is result1 included here? That reward was earned in the first epoch, not now.

A key impl distinction to draw out here is whether the actual gas charged is equal to the gas limit, or is equal to the amount consumed up to but excluding the return value (which is smaller than the gas limit). Expressing that explicitly in the test would help, by comparing the gas used with the gas limit (go-filecoin will consume the full gas limit in this case).

suites/tipset/rewards_penalties.go
anorth: With more investigation is looks a bit harder because we need peers of `WithBLSMessageAndRet` that can not-validate the return value.

suites/tipset/rewards_penalties.go
frrist: I did, corrected.

suites/tipset/rewards_penalties.go
frrist: While I agree this does add some noise, I think commenting why `retval` is needed and stating that its not part of the actual tests is a fair tradeoff right now.

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,64 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
+		td.ExeCtx.Epoch++
+
+		// get the gas used
+		gasUnitsUsed := result1.Receipts[0].GasUsed
+		gasLimit := int64(gasUnitsUsed) - 1
+		// apply the message again with a reduced gasLimit
+		result2 := tb.WithBlockBuilder(
+			drivers.NewBlockBuilder(td, td.ExeCtx.Miner).
+				WithBLSMessageAndCode(td.MessageProducer.MinerControlAddresses(miner, alice,
+					nil, chain.Nonce(1), chain.GasPrice(1), chain.GasLimit(gasLimit)),
```
anorth: Is `chain.GasPrice(1)` redundant here?

anorth (commented):
Great stuff

anorth (approved):

suites/tipset/rewards_penalties.go
```diff
@@ -208,6 +209,64 @@ func TestMinerRewardsAndPenalties(t *testing.T, factory state.Factories) {
...
+
+		// ensure the actual gas charged is equal to the gas limit rather than the amount consumed up to but excluding
+		// the return value which is smaller than the gas limit.
+		assert.Equal(t, gasLimit, int64(result2.Receipts[0].GasUsed))
+
+		thisRwd := big.Add(newRewards2.LastPerEpochReward, result2.Receipts[0].GasUsed.Big())
+		prevRwd := big.Add(prevRewards2.LastPerEpochReward, result1.Receipts[0].GasUsed.Big())
+
+		assert.Equal(td.T, big.Add(prevRwd, thisRwd), td.GetBalance(miner))
+		assert.Equal(td.T, big.Sub(prevRewards2.Treasury, thisRwd), newRewards2.Treasury)
```
anorth: This isn't quite right. The gas cost isn't subtracted from the reward actor, only the block reward.

anorth (approved):
Now that underlying HAMT changes have stabilized, I'll land this (with minor fix)

PR #177, opened 2020-04-28
Update gas costs with message version changes

arajasek:
Values look as expected, save for 342 -> 350 and 168 -> 176


drivers/test.go
```diff
@@ -295,7 +295,7 @@ func (td *TestDriver) Complete() {
 func (td *TestDriver) ApplyMessage(msg *types.Message) (result types.ApplyMessageResult) {
 	defer func() {
 		if r := recover(); r != nil {
-			result.Receipt.ExitCode = exitcode.SysErrInternal
+			result.Receipt.ExitCode = exitcode.SysErrForbidden
```
anorth, line 298: Setting any exit code here at all is misleading. @frrist can we not?

If we must set one, maybe use SysErrReserved1 so that it can't possibly be confused with a legitimate code.

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/filecoin-project/go-address v0.0.2-0.20200218010043-eb9bb40ed5be
 	github.com/filecoin-project/go-crypto v0.0.0-20191218222705-effae4ea9f03
 	github.com/filecoin-project/go-fil-commcid v0.0.0-20200208005934-2b8bd03caca5
-	github.com/filecoin-project/specs-actors v0.2.0
+	github.com/filecoin-project/specs-actors v0.2.1-0.20200428232403-f0282340f59a
```
anorth, line 11: Please don't do this in this PR, do it in a different one that's only about actor upgrades. 

I'm a bit torn about whether to object pointing to a non-released version at all. This increases integration costs all round. There's another release coming which lotus, gfc, and chain validation will all need to consume. Doing this cycle lots of times is a time suck.

anorth (commented):
I'm fine with the gas changes from the message version, and indeed would like to consume that ASAP to avoid disabling chain-validation tests in gfc when I do the same thing.

Please decouple it from pointing at an arbitrary non-release actor version, which gfc is not prepared to consume.

drivers/test.go
frrist: Until we exercise test we expect to panic the `t.Fatal` ought to be enough

box/blob.go
```diff
@@ -9,50 +9,50 @@ import (
 )
 
 func init() {
-	resources.Add("/TestChainValidationMessageSuiteTestAccountActorCreationfailcreateBLSaccountactorinsufficientbalance", []types.ApplyMessageResult{types.ApplyMessageResult{Receipt: types.MessageReceipt{ExitCode: 2, ReturnValue: []uint8{}, GasUsed: 0}, Penalty: abi.NewTokenAmount(178), Reward: abi.NewTokenAmount(0), Root: "bafy2bzacea7fphntvesdry5ohpz6fd46erq3ea7wx3oimynir73lqjhg7tcc4"}})
```
frrist, line 12: yaaay someone that's not me knows how to do this! :bus:

arajasek: hee hee

frrist (approved):

frrist:
I believe you'll need to regenerate gas values and state roots after the rebase since they were created before the message version addition, I think.

arajasek:
Done in #178 and #179

PR #178, opened 2020-04-29
Update gas values and state for messages with version numbers

anorth:


arajasek:
Hmm, just based on a quick eye-ball, the values here seem different than those in #177 (which means Lotus will probably fail them)...Doesn't mean these are wrong, but something's not aligned.

arajasek (approved):
LGTM

PR #180, opened 2020-04-29
Fix ApplyMessage to correctly validate gas/state values

frrist:
Nested sends tests were using the wrongly exported `ApplyMessage` method on the state driver. This caused the validation of their stateroots and gas values to be skipped. The PR corrects the usage, unexports the non-validating-applymessage method, and generates the resources.


drivers/constants.go
```diff
@@ -11,5 +11,5 @@ const (
 
 var (
 	TotalNetworkBalance = big_spec.Mul(big_spec.NewInt(totalFilecoin), big_spec.NewInt(filecoinPrecision))
-	EmptyReturnValue = []byte{}
+	EmptyReturnValue    = []byte{}
```
frrist, line 14: gofmt

drivers/test.go
```diff
@@ -319,13 +305,27 @@ func (td *TestDriver) ApplyFailure(msg *types.Message, code exitcode.ExitCode) t
 }
 
 func (td *TestDriver) applyMessageExpectCodeAndReturn(msg *types.Message, code exitcode.ExitCode, retval []byte) types.ApplyMessageResult {
-	result := td.ApplyMessage(msg)
+	result := td.applyMessage(msg)
 	if !td.validateResult(result, code, retval) {
```
frrist: `ApplyMessage` doesn't perform any validation, it just records the values. This method should not be used by any tests so unexport it.

suites/message/nested.go
```diff
@@ -363,7 +352,8 @@ func (s *msStage) send(to address.Address, value abi.TokenAmount, method abi.Met
 		Params: buf.Bytes(),
 	}
 	msg := s.driver.MessageProducer.MultisigPropose(s.msAddr, s.creator, &pparams, chain.Nonce(approverNonce))
-	return s.driver.ApplyMessage(msg)
+	expRet := typegen.CborInt(0)
+	return s.driver.ApplyExpect(msg, chain.MustSerialize(&expRet))
```
frrist: Multisig propose returns a cbor-int corresponding to the transaction ID, expect that here.

suites/message/nested.go
```diff
@@ -334,23 +320,26 @@ type msStage struct {
...
-	var ret init_.ExecReturn
-	err := ret.UnmarshalCBOR(bytes.NewReader(result.Receipt.ReturnValue))
+	var initAct init_.State
+	td.GetActorState(builtin.InitActorAddr, &initAct)
+	addr, err := address.NewIDAddress(uint64(initAct.NextID))
 	require.NoError(td.T, err)
 
+	ret := td.ComputeInitActorExecReturn(creatorPK, 0, 0, addr)
+
+	td.ApplyExpect(msg, chain.MustSerialize(&ret))
```
anorth: This is convoluted. There should be a method with the same functionality as ApplyMessage that just applies a message and returns the result, without necessarily making assertions about that result. I think I exported ApplyMessage (before gas/state validation was implemented) for this purpose.

suites/message/nested.go
anorth: Please instead expose something that just returns the value, instead of expecting it.

anorth (commented):
Good find and fix, but please expose something equivalent to ApplyMessage (that does validation properly) too.

suites/message/nested.go
frrist: That method would just make assertions about the gas used and resulting state root of the application then?

anorth: Yes. The important thing is to provide the return value for use later in the test.

anorth:
The original commit in this PR regenerated gas values (presumably from lotus) for  some cases that were incorrectly skipping validation before. However from go-filecoin's point of view the existing values are correct - they had not drifted out of date.

There's a divergence here, so pending working out which is the desired behaviour, I've reverted to the existing values.

anorth:
I think the tipset evaluator has the same problem.

PR #183, opened 2020-04-30
Clean up tipset builder validation in Apply

anorth:
Same fix as #180.

Nothing actually calls the bare `Apply()` right now though.


frrist (approved):
:+1:

PR #184, opened 2020-04-30
Update expected CIDs to match new hamt library

arajasek:
Uses commit `f3547695a9a2` of go-hamt-ipld


frrist:
Which impl was used to generate the new values?

arajasek:
> Which impl was used to generate the new values?

Lotus.

whyrusleeping (approved):

PR #187, opened 2020-05-01
test: show difference between bls & secp gas cost

frrist:
resource generated from go-filecoin@ d79111acc7396ebc64128238128033b0173c047c


suites/tipset/messages_application.go
```diff
@@ -4,14 +4,55 @@ import (
...
+				WithSECPMessageOk(
+					td.MessageProducer.Transfer(receiverSECP, senderSECP, chain.Nonce(0), chain.Value(transferAmnt))),
+		).ApplyAndValidate()
+
+		require.Equal(t, 2, len(results.Receipts))
+
+		blsGasUsed := results.Receipts[0].GasUsed
+		secpGasUsed := results.Receipts[1].GasUsed
+		assert.Equal(t, types.GasUnits(246), blsGasUsed)
+		assert.Equal(t, types.GasUnits(268), secpGasUsed)
```
anorth: I think it would be sufficient for this test to check just that SECP > BLS, to reduce the maintenance burden.

anorth (approved):

PR #188, opened 2020-05-01
only use chainval test names in resource path

### LINKED ISSUE #182: Change box resource names not to include string from method not part of chain validation suite
In the resource name `TestChainValidationMessageSuiteTestAccountActorCreationfailcreateBLSaccountactorinsufficientbalance` the part `TestChainValidationMessageSuite` _does not appear in the chain validation repo_ and is generally not under control of the validation suite. This happens to be working only because the test runner methods in Lotus and Go-filecoin share the same exact name.

This dependency is backwards and will lead to fragility. For example, in order to investigate behaviour of a specific test, I create a method:

```
func TestX(t *testing.T) {
	f := NewFactories(&ValidationConfig{
		trackGas:         true,
		checkExitCode:    true,
		checkReturnValue: true,
		checkStateRoot:   true,
	})
	message.TestNestedSends(t, f)
}
```

But this invalidates the resource name and so cannot be used for checking gas/state values.

frrist:
closes #182 

This requires that chain-validation tests must either start with `MessageTest` or `TipSetTest` so that `box/gen.go` can generate their resources correctly.


anorth (approved):
Thanks.

Please add a someday-maybe issue to find a better pattern for the resource generation. Maybe using the package name rather than test name?

PR #191, opened 2020-05-04
pull in puppet actor and add internal send test

frrist:
### Changes 

- update to latest specs-actors to get puppet actor
- add code gen helpers for puppet actor
- add pupet actor message producer methods
- implement simple inner send test using puppet actor failing for due to
lack of balance

### Implementations 
https://github.com/filecoin-project/go-filecoin/pull/4072
https://github.com/filecoin-project/lotus/pull/1675

### DIFF
--- a/drivers/test.go
+++ b/drivers/test.go
@@ -220,8 +220,8 @@ type ActorState struct {
 	State   runtime_spec.CBORMarshaler
 }
 
-func (b *TestDriverBuilder) WithActorState(acts []ActorState) *TestDriverBuilder {
-	b.actorStates = acts
+func (b *TestDriverBuilder) WithActorState(acts ...ActorState) *TestDriverBuilder {
+	b.actorStates = append(b.actorStates, acts...)
 	return b
 }
 
--- a/suites/message/nested.go
+++ b/suites/message/nested.go
@@ -6,6 +6,10 @@ import (
 	"testing"
 
 	"github.com/filecoin-project/go-address"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	typegen "github.com/whyrusleeping/cbor-gen"
+
 	"github.com/filecoin-project/specs-actors/actors/abi"
 	"github.com/filecoin-project/specs-actors/actors/abi/big"
 	builtin "github.com/filecoin-project/specs-actors/actors/builtin"
@@ -13,12 +17,10 @@ import (
 	"github.com/filecoin-project/specs-actors/actors/builtin/multisig"
 	"github.com/filecoin-project/specs-actors/actors/builtin/paych"
 	"github.com/filecoin-project/specs-actors/actors/builtin/reward"
+	"github.com/filecoin-project/specs-actors/actors/puppet"
 	"github.com/filecoin-project/specs-actors/actors/runtime"
 	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
 	"github.com/filecoin-project/specs-actors/actors/util/adt"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	typegen "github.com/whyrusleeping/cbor-gen"
 
 	"github.com/filecoin-project/chain-validation/chain"
 	"github.com/filecoin-project/chain-validation/chain/types"
@@ -26,6 +28,17 @@ import (
 	"github.com/filecoin-project/chain-validation/state"
 	"github.com/filecoin-project/chain-validation/suites/utils"
 )
+
+var PuppetAddress address.Address
+
+func init() {
+	var err error
+	// the address before the burnt funds address
+	PuppetAddress, err = address.NewIDAddress(builtin.FirstNonSingletonActorId - 2)
+	if err != nil {
+		panic(err)
+	}
+}
 
 // Tests exercising messages sent internally from one actor to another.
 // These use a multisig actor with approvers=1 as a convenient staging ground for arbitrary internal messages.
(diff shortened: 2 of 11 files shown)


drivers/test.go
```diff
@@ -8,27 +8,29 @@ import (
...
+
 	abi_spec "github.com/filecoin-project/specs-actors/actors/abi"
 	big_spec "github.com/filecoin-project/specs-actors/actors/abi/big"
 	builtin_spec "github.com/filecoin-project/specs-actors/actors/builtin"
 	account_spec "github.com/filecoin-project/specs-actors/actors/builtin/account"
 	cron_spec "github.com/filecoin-project/specs-actors/actors/builtin/cron"
 	init_spec "github.com/filecoin-project/specs-actors/actors/builtin/init"
 	market_spec "github.com/filecoin-project/specs-actors/actors/builtin/market"
-	"github.com/filecoin-project/specs-actors/actors/builtin/miner"
+	miner_spec "github.com/filecoin-project/specs-actors/actors/builtin/miner"
```
frrist: changed for consistency

anorth: As a general comment: most of the imports in this project are from specs-actors. Name collisions don't look too difficult to avoid, `miner` was the most likely, if anything (and `init_` to avoid the Go reserved word). This convention will be difficult to enforce if many people contribute to this project. It won't last.

I suggest changing the convention so that the easy, tool-supported thing is the default (in a different PR).

drivers/test.go
```diff
@@ -144,6 +151,18 @@ func init() {
...
+	PuppetActorAddress, err = address.NewIDAddress(98)
+	if err != nil {
+		panic(err)
+	}
+	PuppetActorState = ActorState{
+		Addr:    PuppetActorAddress,
+		Balance: abi_spec.NewTokenAmount(0),
+		Code:    puppet_spec.PuppetActorCodeID,
+		State:   &puppet_spec.State{},
+	}
```
anorth: I can't see why the puppet actor is treated differently to the other ones. Please move this code up to 144 and follow the same pattern, including adding to DefaultBuiltinActorsState.

Alternatively, don't put any code here at all: defer it all to the caller of the builder. E.g. maybe they want a puppet with a balance? Maybe they want two puppets? This flexibility could be valuable tho we don't strictly need it now.

Either way is ok, but go all-in one or the other: builtin or not builtin.

suites/message/puppet.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/filecoin-project/specs-actors/actors/builtin"
+	"github.com/filecoin-project/specs-actors/actors/puppet"
+	"github.com/filecoin-project/specs-actors/actors/runtime/exitcode"
+)
+
+func MessageTest_PuppetActorInnerSends(t *testing.T, factories state.Factories) {
+	builder := drivers.NewBuilder(context.Background(), factories).
+		WithDefaultGasLimit(1_000_000).
+		WithDefaultGasPrice(big_spec.NewInt(1)).
+		WithActorState(append(drivers.DefaultBuiltinActorsState, drivers.PuppetActorState))
```
anorth: This would be more ergonomic if `WithActorState` _added_ an actor to the builtin set rather than replacing the builtin set (which requires the caller to do an `append`, which they will always do).

I don't quite understand how this works without requiring the puppet actor's address as a parameter specifying where to put the state. But something like:
```
	.WithActor(myPuppetAddr, myPuppetCodeID, puppet.ConstructState(), myPuppetBalance)
	.WithActor(anotherAddr, someCodeId, foo.ConstructState(), anotherBalance)
```
would make sense.

suites/message/puppet.go
```diff
@@ -0,0 +1,40 @@
...
+
+		alice, _ := td.NewAccountActor(drivers.SECP, actBal)
+		bob, _ := td.NewAccountActor(drivers.SECP, big_spec.Zero())
+
+		td.ApplyFailure(td.MessageProducer.PuppetSend(drivers.PuppetActorAddress, alice, &puppet.SendParams{
+			To:     bob,
+			Value:  transferAmnt,
+			Method: builtin.MethodSend,
+			Params: nil,
+			// TODO: what should this be? Lotus returns SysErrSenderInvalid.
```
anorth: Assuming that everything is set up correctly and `alice` is a valid sender, this should indeed be SysErrInsufficientFunds, and alice should pay gas for the failure.

`SysErrSenderInvalid` would imply the miner is at fault and be penalised, which would be the case if the top-level sender, alice, was not capable of sending the message.

anorth (changes requested):
My hope for this is to replace the use of the multisig actor in `nested.go` rather than produce another set of tests. It's good to start out with just one to demonstrate and communicate the pattern, though. So I'd suggest either:
- move this test into the `nested` file now, or
- hold off this test and replace one of the existing ones (which both impls should still pass), and put the new one in a new PR

We have to be careful about landing incompatibilities between actors version. I think this means that this can't land until the impls have upgraded to a release that includes the puppet actor. Otherwise, if we need either (a) some other test urgently without actor changes, or (b) an urgent actor fix without test changes, we'll be forced to disable or branch chain-validation tests.

This blockage can be resolved with a branching strategy, so the impls can always consume something compatible with the actors release they want. FYI @icorderi

frrist:
> My hope for this is to replace the use of the multisig actor in nested.go rather than produce another set of tests. 

Does this mean replacing _all_ usage of the multisig actor? There are a couple tests where the multisig actors state is inspected -- would said inspections be removed? Edit: *in nested.go

suites/message/nested.go
```diff
@@ -307,6 +320,34 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
 		td.AssertHead(builtin.InitActorAddr, prevHead)  // Init state unchanged.
 	})
 
+	t.Run("fail insufficient funds for transfer in inner send", func(t *testing.T) {
+		puppetBalance := big.Zero()
+		td := builder.WithActorState(drivers.ActorState{
```
frrist, line 325: instantiate the puppet actor as needed for testing to keep stateroots consistent.

anorth:
> Does this mean replacing all usage of the multisig actor? 

Probably not. There are some other tests that use it just as a convenient actor with state to mutate and inspect. But all of nested.go, probably.

suites/message/nested.go
anorth: Great. When this drivers.ActorState construction appears more than once please extract to a function.

anorth (approved):
Looks good. Please hold off - there's a new actor release coming soon.

PR #193, opened 2020-05-05
test lack of gas to cover actor creation

### LINKED ISSUE #192: test insufficient gas to create new account actor
From Slack Conversation:
```
Why  4:27 PM:
Interop issue we just hit and would be nice to have a 
test for: Not having enough gas to create a new account
actor through a simple send
```
```
Why  4:27
gfc returns exit code 7, and we return exit code 1 i think
```

frrist:
This test would be more robust if chain-validation had a price-list. This would allow the test to be written in a way s.t. it could ensure the gasLimit in the message was just enough to cause its application to fail when the runtime checks if there is enough gas to create a new actor. Currently the test relies on a handcrafted gasLimit (200) to cause it to fail at the right spot.

closes #192


suites/message/message_application.go
```diff
@@ -46,13 +46,26 @@ func MessageTest_MessageApplicationEdgecases(t *testing.T, factory state.Factori
 			td.MessageProducer.Transfer(alice, alice, chain.Value(transferAmnt), chain.Nonce(0), chain.GasPrice(10), chain.GasLimit(1)),
 			exitcode.SysErrOutOfGas)
 
-		// Expect Message application to fail due to lack of gas when sender address is unknown
+		// Expect Message application to fail due to lack of gas when the message being sent has invalid destination.
```
anorth: I think the original comment is correct. I find the argument order of `Transfer` to be very confusing though so had to go inspect it to confirm this.

suites/message/message_application.go
```diff
@@ -46,13 +46,26 @@ func MessageTest_MessageApplicationEdgecases(t *testing.T, factory state.Factori
...
+	t.Run("fail not enough gas to cover account actor creation", func(t *testing.T) {
+		td := builder.Build(t)
+		defer td.Complete()
+
+		alice, _ := td.NewAccountActor(drivers.SECP, aliceBal)
+		newAccount := utils.NewSECP256K1Addr(t, "1")
+
+		// Expect Message application to fail due to lack of gas
+		td.ApplyFailure(
+			td.MessageProducer.Transfer(newAccount, alice, chain.Value(transferAmnt), chain.Nonce(0), chain.GasPrice(1), chain.GasLimit(200)),
```
anorth: I think we can do better. How about:
- Start by finding the true gas cost by sending a successful message.
- Then try every gas value mod `something` less than the true cost, and check we get the same exit code

`something` could probably be as large as 50 and still very robust to the gas table changing (since the true cost will be >500).

suites/message/message_application.go
frrist: Filed https://github.com/filecoin-project/chain-validation/issues/197 to address parameter ordering.

suites/message/message_application.go
```diff
@@ -46,13 +46,41 @@ func MessageTest_MessageApplicationEdgecases(t *testing.T, factory state.Factori
...
+
+		alice, _ := td.NewAccountActor(drivers.SECP, aliceBal)
+		aliceNonce := uint64(0)
+		aliceNonceF := func() uint64 {
+			defer func() { aliceNonce++ }()
+			return aliceNonce
+		}
+		newAccountA := utils.NewSECP256K1Addr(t, "1")
+
+		// get the "true" gas cost of applying the message
```
frrist, line 68: Note: the true gas cost cannot be determined by applying a message, this is an approximation. Chain-validation uses a default gas limit of [`1,000,000`](https://github.com/filecoin-project/chain-validation/pull/193/files#diff-fdc8ab8a15ed13b6f5c5724adf12dadeL22) and a default gas price of [`1`](https://github.com/filecoin-project/chain-validation/pull/193/files#diff-fdc8ab8a15ed13b6f5c5724adf12dadeR23). These values are encoded into a message and since they are large they increase the size of the message on chain and thus its cost.

anorth (approved):

PR #194, opened 2020-05-05
pull the resource key from the stack

frrist:
### Motivation

Allow one-off chain-validation tests to be called without invoking their suite while preserving access to their resources. e.g.

```go
func TestSomethingRealQuick(t *testing.T) {
	f := NewFactory()
	message.MessageTest_MessageApplicationEdgecases(t, f)
}
```

### DIFF
--- a/tracker/track.go
+++ b/tracker/track.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -152,7 +153,22 @@ func filenameFromTest(t testing.TB) string {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// this is a really hacky way of getting the chain-validation test method name.
+	pc := make([]uintptr, 6)
+	n := runtime.Callers(5, pc)
+	frames := runtime.CallersFrames(pc[:n])
+	frame, _ := frames.Next()
+	frameTokens := strings.Split(frame.Function, ".")
+	parentTestMethodName := frameTokens[2]
+
+	var childTestMethodName string
 	tokens := strings.Split(t.Name(), "/")
-	chainvalTestName := strings.Join(tokens[1:], "")
-	return fmt.Sprintf("/%s", reg.ReplaceAllString(chainvalTestName, ""))
+	// if the parent test doesn't have children
+	if len(tokens) == 1 {
+		childTestMethodName = ""
+	} else {
+		childTestMethodName = tokens[len(tokens)-1]
+	}
+	testFileName := fmt.Sprintf("%s%s", parentTestMethodName, childTestMethodName)
+	return fmt.Sprintf("/%s", reg.ReplaceAllString(testFileName, ""))
 }


tracker/track.go
```diff
@@ -152,7 +153,22 @@ func filenameFromTest(t testing.TB) string {
...
 	tokens := strings.Split(t.Name(), "/")
-	chainvalTestName := strings.Join(tokens[1:], "")
-	return fmt.Sprintf("/%s", reg.ReplaceAllString(chainvalTestName, ""))
+	// if the parent test doesn't have children
+	if len(tokens) == 1 {
+		childTestMethodName = ""
+	} else {
+		childTestMethodName = tokens[len(tokens)-1]
+	}
+	testFileName := fmt.Sprintf("%s%s", parentTestMethodName, childTestMethodName)
```
frrist, line 172: During normal usage tokens will contain one of the following:

- Parent test has children tests, i.e. calls `t.Run`
```
tokens =
 0 = {string} "TestChainValidationTipSetSuite"                      // test name from go-filecoin or lotus
 1 = {string} "TipSetTest_BlockMessageApplication"                  // (parent) test name defined in chain-validation
 2 = {string} "SECP_and_BLS_messages_cost_different_amounts_of_gas" // (child) test name
```
- Parent test has no children
```
tokens =
 0 = {string} "TestChainValidationMessageSuite"
 1 = {string} "MessageTest_InitActorSequentialIDAddressCreate"
```

During one-off usage tokens will contain one of the following:
- Parent tests has children
```
tokens =
 0 = {string} "TestSomethingRealQuick"
 1 = {string} "success_create_SECP256K1_account_actor"
```
- Parent test has no children
```
tokens = 
 0 = {string} "TestSomethingRealQuick"
```

Walking the stack is the only way (given the current testharness - `*testing.T`) to get the parent tests name.

anorth:
This change is way more complicated than I expect. It suggests to me that the goals are not clear, or there are perceived constraints that may not be real constraints. It's also of course possible that I don't understand the constraints.

For example, I do not think there is a particular need for the top level test name to be used to uniqueify things - it won't be hard to maintain unique case names across different suites. Is that what you're attempting here?

frrist:
I am trying to enable a developer to call a one-off chain-validation test from gfc or lotus while preserving access to the tests resource file. 

The constraint is chain-validation tests aren't actually go tests (they accept more than a `*testing.T`), meaning `t.Name()` doesn't return their name, it returns the name of whatever the developer called their test. For example, if a developer did the following:
```go
func TestSomethingRealQuick(t *testing.T) {
  f := NewFactory()		
  message.MessageTest_AccountActorCreation(t, f)
}
```
`t.Name()` ([called here](https://github.com/filecoin-project/chain-validation/blob/master/tracker/track.go#L149)) will return `"TestSomethingRealQuick"` and **not** `"MessageTest_AccountActorCreation"` (the expected resource key). This works for the normal case since tests are called like:
```go
f := factory.NewFactories()
for _, testCase := range suites.TipSetTestCases() {
  if TestSuiteSkipper.Skip(testCase) {
    continue
  }
  t.Run(caseName(testCase), func(t *testing.T) {
    testCase(t, f)
  })
}

```

Keying resource names on `t.Name()` is an annoying constraint to work around. A simpler alternative could be implementing our own testing harness that better facilitates the resource access we are performing here. Sadly this isn't as simple as implementing the `testing.TB` interface and plumbing it through as Go [intentionally prevents this](https://golang.org/src/testing/testing.go#L567) :disappointed:.

anorth:
> Keying resource names on t.Name() is an annoying constraint to work around.

Right. I don't think it need be a constraint. But we don't need to implement a whole testing harness to avoid it. Can we just use some string constants explicitly written into the code? They don't necessarily need to be hierarchical - collisions will probably be easy to detect and avoid.

PR #203, opened 2020-05-13
add undecodable params during inner send

frrist:
- uses puppet actor
- doesn't have resources generated yet since both lotus and go-filecoin fail (need conformation this test is correct)

Lotus fails applying both messages with error code: `SysErrInsufficientFunds`, it never reaches the internal send. This appears to be wrong.
https://github.com/filecoin-project/lotus/blob/master/chain/vm/runtime.go#L352

go-filecoin passes the first message application and panics on applying the second message.

### DIFF
--- a/suites/message/nested.go
+++ b/suites/message/nested.go
@@ -29,17 +29,6 @@ import (
 	"github.com/filecoin-project/chain-validation/suites/utils"
 )
 
-var PuppetAddress address.Address
-
-func init() {
-	var err error
-	// the address before the burnt funds address
-	PuppetAddress, err = address.NewIDAddress(builtin.FirstNonSingletonActorId - 2)
-	if err != nil {
-		panic(err)
-	}
-}
-
 // Tests exercising messages sent internally from one actor to another.
 // These use a multisig actor with approvers=1 as a convenient staging ground for arbitrary internal messages.
 func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
@@ -323,12 +312,7 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
 	t.Run("fail insufficient funds for transfer in inner send", func(t *testing.T) {
 		// puppet actor has zero funds
 		puppetBalance := big.Zero()
-		td := builder.WithActorState(drivers.ActorState{
-			Addr:    PuppetAddress,
-			Balance: puppetBalance,
-			Code:    puppet.PuppetActorCodeID,
-			State:   &puppet.State{},
-		}).Build(t)
+		td := builder.WithActorState(drivers.MakePuppetActorState(puppetBalance)).Build(t)
 		defer td.Complete()
 
 		alice, _ := td.NewAccountActor(drivers.SECP, acctDefaultBalance)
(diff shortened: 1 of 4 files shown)


suites/message/nested.go
```diff
@@ -359,6 +359,65 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
...
+			State:   &puppet.State{},
+		}).Build(t)
+		defer td.Complete()
+
+		// user separate actors for easier balance assertions
+		alice, _ := td.NewAccountActor(drivers.SECP, acctDefaultBalance)
+		bob, _ := td.NewAccountActor(drivers.SECP, acctDefaultBalance)
+
+		result1 := td.ApplyMessage(
+			td.MessageProducer.PuppetSendMarshalCBORFailure(alice, PuppetAddress, &puppet.SendParams{
```
frrist: cause the puppet actor to invoke an inner send with undecodable params: https://github.com/filecoin-project/specs-actors/blob/master/actors/puppet/puppet.go#L75

suites/message/nested.go
```diff
@@ -359,6 +359,65 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
 
 	})
 
+	t.Run("fail undecodable params nested send", func(t *testing.T) {
+		puppetBalance := big.NewInt(1_000)
+		td := builder.WithActorState(drivers.ActorState{
+			Addr:    PuppetAddress,
+			Balance: puppetBalance,
+			Code:    puppet.PuppetActorCodeID,
+			State:   &puppet.State{},
```
anorth: Factor this repeated state construction out to a function.

suites/message/nested.go
```diff
@@ -359,6 +359,65 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
...
+		// outer send should pass
+		assert.Equal(t, result1.Receipt.ExitCode, exitcode.Ok)
+
+		// prevent a panic in test code
+		require.NotEmpty(t, result1.Receipt.ReturnValue)
+		var puppetRet puppet.SendReturn
+		chain.MustDeserialize(result1.Receipt.ReturnValue, &puppetRet)
+
+		// the inner message must fail
+		assert.Equal(t, exitcode.SysErrInvalidParameters, puppetRet.Code)
```
anorth: I expect `ErrSerialization` as this exit code. It's logically since the actor code.

`SysErrInvalidParameters` is for when parameters cannot be *de*coded into the expected shape (which is [distinct thing to test](https://github.com/filecoin-project/chain-validation/blob/483332336e8ecf9982c8bbeabd4796856c8d0aea/suites/message/message_application.go#L178), by crafting a message with the empty/wrong parameters for the method being called).

@magik6k we should consider changing this to also use `ErrSerialization` (and remove SysErrInvalidParameters completely) to logically have the de-serialization be in the actor code, following our earlier decisions about store Get/Put errors. If you agree, we can change this test now and the update impls to match.

suites/message/nested.go
```diff
@@ -359,6 +359,65 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
...
+
+		// outer send should pass
+		assert.Equal(t, result2.Receipt.ExitCode, exitcode.Ok)
+
+		// prevent a panic in test code
+		require.NotEmpty(t, result2.Receipt.ReturnValue)
+		chain.MustDeserialize(result2.Receipt.ReturnValue, &puppetRet)
+
+		// the inner message must fail
+		assert.Equal(t, exitcode.SysErrInvalidParameters, puppetRet.Code)
```
anorth: Ditto

anorth (approved):

frrist:
Status Update:
The PR is awaiting Lotus to update to specs-actors 0.5.5 and for a fix in lotus so that the resources can be generated for the new test added.

PR #205, opened 2020-05-14
add test for missing message recipient

laser:
## What's in this PR?

This changeset adds a simple test case affirming the system's behavior when a message is sent to an address corresponding to an actor which does not yet exist.


suites/message/message_application.go
```diff
@@ -141,8 +141,20 @@ func MessageTest_MessageApplicationEdgecases(t *testing.T, factory state.Factori
...
+		defer td.Complete()
+
+		alice, _ := td.NewAccountActor(drivers.SECP, aliceBal)
+
+		// Sending a message to non-existent actor must produce an error.
+		unknown := utils.NewIDAddr(t, 10000000)
+		td.ApplyFailure(
+			td.MessageProducer.Transfer(alice, unknown, chain.Value(transferAmnt), chain.Nonce(0)),
+			exitcode.SysErrInvalidReceiver)
+	})
```
frrist, line 175: This test can be extended to also test sending to a non-existent _actor_ address (`utils.NewActorAddr`) in addition to the ID address tested here.

frrist (approved):

anorth (approved):

suites/message/message_application.go
laser: Done!

frrist (approved):

PR #206, opened 2020-05-14
assert proper error code when method doesn't exist for recipient actor

laser:
## What's in this PR?

This new test asserts the error code produced when a recipient is sent a message which uses a method that it does not support.


frrist (approved):
b-e-a-utiful

anorth (approved):

Issue #169, opened 2020-04-16
test consensus slashing


whyrusleeping:
try for decent line coverage on all the cases, should be pretty easily doable with just faking blocks

PR #208, opened 2020-05-15
add syscalls to testdriver and make funcs swapable

frrist:
### Motivation
in order to fulfill https://github.com/filecoin-project/chain-validation/issues/169 (consensus slashing) chain-validation tests need control over what `VerifyConsensusFault` returns.

This change allows the syscall method implementations to be configured from within a tests, e.g.:
```go
func TestSomeConssnsusThingy() {
// test setup

		td.SysCalls.VerifyConsensusFaultFunc = func(h1, h2, extra []byte) (*runtime.ConsensusFault, error) {
			return &runtime.ConsensusFault{
				Target: address.Address{},
				Epoch:  0,
				Type:   0,
			}, nil

		}

// assertions on fault handling
}
```

This is an API breaking change so expect changes to lotus and go-filecoin to follow.


anorth:
Looks good, but I want to confirm where this is going.

There are two different things potentially in scope for this:
- testing that actor code behaves correctly when the syscall returns fault or no fault
- testing that the fault detection logic embedded in each impl behaves correctly

Chain validation is most suited for the _second_ one, and we need comprehensive coverage. That means we do need the impl to provide their syscalls implementation for use in tests. The _first_ one is better addressed by actor unit tests, which already have mocking facilities and depend not on the runtime implementation.

#169 is ambiguous as to which cases it's talking about covering. I interpret it as all the cases by which two block headers might or might not constitute a fault.

anorth (approved):
Looks good, but please don't land this on master until both impls are ready to consume it (i.e. have a branch demonstrating it), lest it block propagation of an urgent actor/validation change.

frrist:
Thanks. I was unsure of the preferred approach for testing consensus faults (mocking or "for-reals"), this PR works with either. I'll move forward by testing with implementation specific syscalls. cc @whyrusleeping for context in case this doesn't align with your expectations from issue #169

frrist:
@anorth PR's in gfc and lotus here:
https://github.com/filecoin-project/go-filecoin/pull/4156
https://github.com/filecoin-project/lotus/pull/1815

Give this comment a :+1: if all looks good and I'll merge this and update the aforementioned PRs with the new version

PR #209, opened 2020-05-28
update to batch verification porep changes

whyrusleeping:
depends on this fix: https://github.com/filecoin-project/specs-actors/pull/401

And also needs the test vectors updated as this changes the power actor and miner state roots


arajasek:
Test vectors updated

frrist (approved):

PR #210, opened 2020-05-28
Add another self transfer test

arajasek:
Demonstrates a bug in Lotus that's fixed in https://github.com/filecoin-project/lotus/pull/1861


suites/message/transfer.go
```diff
@@ -151,6 +151,19 @@ func MessageTest_ValueTransferAdvance(t *testing.T, factory state.Factories) {
 		td.AssertBalance(alice, big_spec.Sub(aliceInitialBalance, result.Receipt.GasUsed.Big()))
 	})
 
+	t.Run("self transfer different address types", func(t *testing.T) {
+		td := builder.Build(t)
+		defer td.Complete()
+
+		alice, aliceId := td.NewAccountActor(drivers.SECP, aliceInitialBalance)
```
frrist, line 158: Can we exercise the rest of the self transfer cases here while were at it?
`IdAddress -> IdAddress`
`IdAddress -> PubKeyAddress`

frrist (approved):

suites/message/transfer.go
arajasek: Donesies

PR #216, opened 2020-07-10
Update to new gas values

Kubuxu:


suites/message/multisig.go
```diff
@@ -12,22 +16,16 @@ import (
...
 
 func MessageTest_MultiSigActor(t *testing.T, factory state.Factories) {
 	builder := drivers.NewBuilder(context.Background(), factory).
-		WithDefaultGasLimit(1_000_000).
+		WithDefaultGasLimit(1_000_000_000).
 		WithDefaultGasPrice(big_spec.NewInt(1)).
 		WithActorState(drivers.DefaultBuiltinActorsState...)
 
 	t.Run("constructor test", func(t *testing.T) {
-		const numApprovals = 3
```
anorth: What prompted this change?

anorth (approved):
Thanks!

suites/message/multisig.go
Kubuxu: @arajasek you made it change ^^

PR #231, opened 2020-07-31
Update cborgen and expected values

arajasek:


suites/message/multisig.go
```diff
@@ -81,7 +81,7 @@ func MessageTest_MultiSigActor(t *testing.T, factory state.Factories) {
 			To:     outsider,
 			Value:  valueSend,
 			Method: builtin_spec.MethodSend,
-			Params: []byte{},
+			Params: nil,
```
arajasek, line 84: Confirm this seems reasonable?

magik6k (approved):

PR #233, opened 2020-08-03
Add tests for sender balance below gas requirement

anorth:
#226 changed the rules to allow a message sender to be unable to cover the value being transferred, letting the failure go on chain and charging the sender gas. However, the change lost coverage for the case when the sender can't even afford the gas for the message failure to go on chain, in which case the miner should be penalized.

Note: being out of touch with Lotus current branches etc I have not yet computed the new state/gas values for these tests. @arajasek would you be so kind as to take this PR over and add them?


arajasek (approved):
Thanks for this! It's ready to land now :)

anorth:
Thank you!

PR #234, opened 2020-08-03
Burn gas from the gas limit in excess of 1.3*gasUsed

arajasek:
The primary changes are:

- Supporting the new idea of burning extra gas supplied in the `gasLimit` beyond a threshold of `1.3 * gasUsed`. Introduced in https://github.com/filecoin-project/lotus/pull/2764/files
- Adds a calculator whose `CalcMessageCost()` indicates the amount that a message sender's balance should decrease by based on the message and its receipt (factoring in gas used, `value` transferred, and excess gas burned).


chain/types/message.go
```diff
@@ -17,16 +17,16 @@ import (
 
 type Message struct {
 	// Address of the receiving actor.
-	To address.Address
+	To string
```
anorth: These seem like a ton of unrelated changes. Not necessarily bad, but what's the motivation?

chain/types/receipt.go
```diff
@@ -7,6 +7,7 @@ import (
 
 // MessageReceipt is the return value of message application.
 type MessageReceipt struct {
+	Msg         Message
```
anorth: It looks like GasLImit and GasPrice are the only relevant properties, so you could reduce unnecessary coupling by only inlining those here. Would that allow revert of all the type changes above (I'm not sure if they were motivated by this).

drivers/test.go
```diff
@@ -446,10 +446,13 @@ func (td *TestDriver) AssertBalance(addr address.Address, expected abi_spec.Toke
 }
 
 // Checks an actor's balance and callSeqNum.
-func (td *TestDriver) AssertActor(addr address.Address, balance abi_spec.TokenAmount, callSeqNum uint64) {
+func (td *TestDriver) AssertActor(addr address.Address, prevBalance abi_spec.TokenAmount, rct types.MessageReceipt, callSeqNum uint64) {
```
anorth: This was definitely intending to assert the balance now, not before. This is getting a bit confusing. 

From looking at the call sites, I think I understand the intention here is to move the expected calculation here. It makes sense when there's no value transfer, as in many cases. It's a little opaque. How about adding a "transferred" parameter here too, so the complete calculation can be inside this method. That would clean up a bunch of cases where you've had to hackily adjust the "prev" balance at the caller including, confusingly, for self-transfers where the net should be zero.

 I think it'll then be fairly clear that the `(prior, transferred, receipt)` params then give the expected post balance.

suites/message/nested.go
```diff
@@ -59,11 +59,9 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
 		stage := prepareStage(td, acctDefaultBalance, multisigBalance)
 		balanceBefore := td.GetBalance(stage.creator)
 
-		// Multisig sends back to the creator.
-		amtSent := abi.NewTokenAmount(1)
-		result := stage.sendOk(stage.creator, amtSent, builtin.MethodSend, nil, nonce)
+		result := stage.sendOk(builtin.BurntFundsActorAddr, abi.NewTokenAmount(1), builtin.MethodSend, nil, nonce)
 
-		td.AssertActor(stage.creator, big.Sub(big.Add(balanceBefore, amtSent), result.Receipt.GasUsed.Big()), nonce+1)
+		td.AssertActor(stage.creator, balanceBefore, result.Receipt, nonce+1)
```
anorth: This is now missing the essential assertion that balance was transferred from the multisig to the destination!

suites/message/nested.go
```diff
@@ -214,8 +212,8 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
 		amtSent := abi.NewTokenAmount(1)
 		result := stage.sendOk(stage.creator, amtSent, abi.MethodNum(99), nil, nonce)
 
-		td.AssertBalance(stage.msAddr, multisigBalance)                                       // No change.
-		td.AssertBalance(stage.creator, big.Sub(balanceBefore, result.Receipt.GasUsed.Big())) // Pay gas, don't receive funds.
+		td.AssertBalance(stage.msAddr, multisigBalance) // No change.
+		td.AssertActor(stage.creator, balanceBefore, result.Receipt, nonce+1)
```
anorth: Retain the comment, please

suites/message/transfer.go
```diff
@@ -122,11 +122,12 @@ func MessageTest_ValueTransferSimple(t *testing.T, factories state.Factories) {
 			// create a message to transfer funds from `to` to `from` for amount `transferAmnt` and apply it to the state tree
 			// assert the actor balances changed as expected, the receiver balance should not change if transfer fails
 			if tc.code.IsSuccess() {
-				td.AssertBalance(tc.sender, big_spec.Sub(big_spec.Sub(tc.senderBal, tc.transferAmnt), result.Receipt.GasUsed.Big()))
+				td.AssertActor(tc.sender, tc.senderBal, result.Receipt, 1)
```
anorth: What about the subtraction of `transferAmnt` from the sender's balance?

suites/message/transfer.go
```diff
@@ -122,11 +122,12 @@ func MessageTest_ValueTransferSimple(t *testing.T, factories state.Factories) {
...
 				td.AssertBalance(tc.receiver, tc.transferAmnt)
 			} else {
 				if tc.code == exitcode.SysErrInsufficientFunds {
-					td.AssertBalance(tc.sender, big_spec.Sub(tc.senderBal, result.Receipt.GasUsed.Big()))
+					// TODO: Hackily re-adding the transfer is bad, consider adding a param indicating whether transfer should be subtracted or not
```
anorth: Isn't that just `code.IsSuccess`? It doesn't matter what the exit code is otherwise, the value should not move.

These changes have definitely confused me.

arajasek: That's...exactly what I need, yeah, thank you so much! (and don't write code late at night).

suites/message/transfer.go
arajasek: That is subsumed into the `AssertActor` method.

suites/message/nested.go
arajasek: Fixed, thanks!

drivers/test.go
arajasek: Done (using the receipt's exitcode, as you suggested).

chain/types/receipt.go
arajasek: Unfortunately, I think we use the To, From, GasLimit, GasPrice, and Value.

I also don't think there's much wrong with embedding the Message into the MessageReceipt (there's good reason not to do so in the implementation itself, but I think it makes sense in a more limited sense such as chain-val).

chain/types/message.go
arajasek: Adding this info in the PR description.

anorth: Thanks for the description.

This is a very unfortunate change on the whole. These structures are intended to match the on-chain structures exactly. This is particularly critical for future integration by other implementations, either through an RPC mechanism or CBOR files, where these are to be exchanged with another impl using essentially the same wire format as the network. I understand now that this change is driven by the inclusion of the `Message` inside the `MessageReceipt`, which is serialized as JSON into the expectation files. I didn't pick up on it the first time round, but altering that structure from its protocol definition is also undesirable for the same reason.

I think we need to find another way to do this. This format change will cripple the integration with other projects, and the ability to export actual chain CAR files using messages build with the builders here.

chain/types/receipt.go
anorth: In my earlier comment I didn't realise that this was the MessageReceipt structure, intended to match the protocol spec structure exactly. I thought this was a higher level "evaluation result" structure, specific to chain validation.

We might need to introduce such a structure to solve the problem here.

Edit: we should not include the Value here.  That makes the tests way too magic and hard to follow. So that one should be excluded anyway on other grounds.

`From` and `To` are only used for some obscure check in the gas calculator that I don't understand. If I understood what it was attempting, we could probably find a better way to reach the same outcome.

drivers/state.go
```diff
@@ -116,6 +116,41 @@ func (d *StateDriver) NewAccountActor(addrType address.Protocol, balanceAttoFil
...
+	}
+	for k, v := range d.actorIDMap {
+		if v == addr {
+			return k
+		}
+	}
+	return address.Undef
+}
+
+func (d *StateDriver) AddrEquals(addr1, addr2 string) bool {
```
anorth: 1. This is more fallout from changing the MessageReceipt structure to be non-canonical, but
2. Why don't you just compare strings?

drivers/test.go
anorth: I'm expecting to see the type signature change to add a `transferred` parameter.

suites/message/paych.go
```diff
@@ -139,14 +139,17 @@ func MessageTest_Paych(t *testing.T, factory state.Factories) {
...
 		// advance the epoch so the funds may be redeemed.
 		td.ExeCtx.Epoch += paych_spec.SettleDelay
 
+		prevBal := td.GetBalance(receiver)
 		collectResult := td.ApplyOk(
 			td.MessageProducer.PaychCollect(receiver, paychAddr, nil, chain.Nonce(1), chain.Value(big_spec.Zero())))
 
 		// receiver_balance = initial_balance + paych_send - settle_paych_msg_gas - collect_paych_msg_gas
-		td.AssertBalance(receiver, big_spec.Sub(big_spec.Sub(big_spec.Add(toSend, initialBal), settleResult.Receipt.GasUsed.Big()), collectResult.Receipt.GasUsed.Big()))
+		td.AssertActor(receiver, big_spec.Add(prevBal, toSend), collectResult.Receipt, 2)
```
anorth: To be clear, I'm expecting/requesting this to be `td.AssertActor(receiver, prevBal, toSend, collectResult.Receipt, 2)`, pushing addition down, so that is not passing `prev + sent` as something called `prevBalance`.

suites/message/transfer.go
anorth: I'm really confused, how can it possibly know the right amount is `transferAmnt`?

Even if it's right, it's too confusing to be a good pattern.

anorth: Ohhh, it's getting it from the "receipt" which includes the whole message. Way too confusing, I think we should not include that value in the receipt or result at all, so that tests must spell it out.

drivers/calculator.go
```diff
@@ -0,0 +1,30 @@
...
+		gb = rct.Msg.GasLimit - overestimate
+	}
+
+	return gb
+}
+
+func (d *StateDriver) CalcMessageCost(rct types.MessageReceipt) big_spec.Int {
+	change := int64(rct.GasUsed) + GetBurn(rct)
+	change = change * rct.Msg.GasPrice
+	if rct.ExitCode.IsSuccess() && !d.AddrEquals(rct.Msg.From, rct.Msg.To) {
```
anorth: What is this address comparison for?

anorth (changes requested):

suites/message/nested.go
```diff
@@ -57,13 +57,15 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
...
+		to, _ := td.NewAccountActor(address.SECP256K1, big.Zero())
 		balanceBefore := td.GetBalance(stage.creator)
 
-		// Multisig sends back to the creator.
 		amtSent := abi.NewTokenAmount(1)
-		result := stage.sendOk(stage.creator, amtSent, builtin.MethodSend, nil, nonce)
+		result := stage.sendOk(to, amtSent, builtin.MethodSend, nil, nonce)
 
-		td.AssertActor(stage.creator, big.Sub(big.Add(balanceBefore, amtSent), result.Receipt.GasUsed.Big()), nonce+1)
+		td.AssertActorChange(stage.creator, balanceBefore, result.Msg.GasLimit, result.Msg.GasPrice, result.Msg.Value, result.Receipt, result.Msg.CallSeqNum+1)
```
anorth: Here and it many of the others you can use `amtSent`, or `big.Zero()` when nothing is sent, instead of `result.Msg.Value`. It's more concise and stronger, enforcing that the message that was processed has the result the test case knows it should.

anorth: Also use the existing `nonce+1` for the same reason: the test is trying to assert that it knows what the nonce is, but you've removed that and exposed it to bugs in the calculation of `Msg.CallSeqNum`.

suites/message/nested.go
```diff
@@ -79,7 +81,7 @@ func MessageTest_NestedSends(t *testing.T, factory state.Factories) {
 		result := stage.sendOk(newAddr, amtSent, builtin.MethodSend, nil, nonce)
 
 		td.AssertBalance(stage.msAddr, big.Sub(multisigBalance, amtSent))
-		td.AssertBalance(stage.creator, big.Sub(balanceBefore, result.Receipt.GasUsed.Big()))
+		td.AssertActorChange(stage.creator, balanceBefore, result.Msg.GasLimit, result.Msg.GasPrice, result.Msg.Value, result.Receipt, result.Msg.CallSeqNum+1)
```
anorth: Here and in many below you've added unnecessary checking of the call sequence number, making this change look more verbose than necessary. Add an `AssertSenderBalance` that's just `AssertBalance` with the gas/value parameters but without nonce. In fact you probably want that almost all the time, and could revert AssertActor.

suites/tipset/rewards_penalties.go
```diff
@@ -45,26 +45,26 @@ func TipSetTest_MinerRewardsAndPenalties(t *testing.T, factory state.Factories)
...
+						WithBLSMessageOk(msg2),
 				).ApplyAndValidate()
 				tipB.Clear()
 
 				td.ExeCtx.Epoch++
 
 				// Each account has paid gas fees.
-				td.AssertBalance(aliceId, big.Sub(aBal, result.Receipts[0].GasUsed.Big()))
-				td.AssertBalance(bobId, big.Sub(bBal, result.Receipts[1].GasUsed.Big()))
+				td.AssertBalance(aliceId, big.Sub(aBal, td.CalcMessageCost(msg1.GasLimit, msg1.GasPrice, big.Zero(), result.Receipts[0])))
```
anorth, line 65: AssertSenderBalance will help here too

anorth (approved):
Thanks. This change as a whole is now half the size, and with much clearer semantics. There's room for more conciseness before landing it.

We could probably follow up with some sugar to make those additional parameters easier to pass, but I don't think that's a big deal.

